PR #1, opened 2019-09-29
Dockerfile

filalex77:


neenjaw:
Awesome, thanks filalex!

can you rebase and squash commits?

neenjaw:
I just thought of something before this is merged. The analyzer only has 10 seconds to run, and maybe we shouldn't be building the escript when the the docker container is started, but instead should be run in a preparation step when the repo is deployed to the the exercism environment so that those 10 seconds are reserved for actual analysis. 

Maybe @NobbZ or @ihid can comment further?

README.md
```diff
@@ -124,7 +124,7 @@ by adding `elixir_analyzer` to your list of dependencies in `mix.exs`:
 ```elixir
 def deps do
   [
-    {:ex_mentor, "~> 0.1.0"}
+    {:elixir_analyzer, "~> 0.1.0"}
```
NobbZ: I think this whole section can get removed, as its unlikely that the analyzer gets published on hex.

.dockerignore
```diff
@@ -0,0 +1 @@
+.git
```
NobbZ: Can you also add `deps` and `_build` as well as variations of the built executable as created by the build script?

NobbZ:
@exercism/ops please review, I can't request a review from you for some reason…

NobbZ:
> Nevermind, this should be cached when the docker image is built. Correct? @NobbZ?

Yes, the escript is prebuilt during building the image. It does not need to be rebuilt as part of the analysis.

neenjaw:
Do we need to wait for @exercism/ops?

NobbZ:
> Do we need to wait for @exercism/ops?

Yes. Every PR touching the `Dockerfile` needs to be reviewed by them:

https://github.com/exercism/automated-analysis/blob/master/docs/docker.md#L9

> - All changes to dockerfiles require a PR review from the @exercism/ops.

neenjaw:
Good to know, I just saw the status changed to approved so wasn't sure if they had reviewed but couldn't find their stamp anywhere

iHiD:
@ccare Could you check this one pls?

PR #3, opened 2019-09-29
Add GitHub Action for Elixir CI

filalex77:
For now, this will just confirm it builds.

As soon as the test suite expands, it will be picked up here.

### DIFF
--- a/.github/workflows/elixir.yml
+++ b/.github/workflows/elixir.yml
@@ -1,0 +1,21 @@
+name: Elixir CI
+
+on: push
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+
+    container:
+      image: elixir:1.8-alpine
+
+    steps:
+    - uses: actions/checkout@v1
+    - name: Install Dependencies
+      run: |
+        mix local.rebar --force
+        mix local.hex --force
+        mix deps.get
+    - name: Run Tests
+      run: mix test


neenjaw:
in PR #4 I am implementing a basic test unit, cleaning up some of the old ways I was testing the prototype.  If you rollback commit `Remove default 'Hello World' test` 788d9db we should be able to merge them both without conflict.

neenjaw:
I'm not familiar enough with GHA to know if this is what we need, but it seems ok to me. @NobbZ I think has more experience, I have tagged him for review.

filalex77:
@NobbZ It fails because there is no corresponding function in source. Either the function should be added, or the test should stay removed

NobbZ:
I haven't actually looked into the test itself, I assumed its the regular `assert true`…

filalex77:
@neenjaw I think this is safe to merge now.

filalex77:
Great, it worked with tests from #8 :tada:
I'd be glad to improve this workflow if anyone has anything to suggest.

neenjaw:
Eventually I think we need to do what the go/JavaScript team is doing with comments. I think they have a build check that any comment in the analyzer must have the matching markup in the [exercise/website-copy](https://github.com/exercism/website-copy/tree/master/automated-comments) repo

PR #10, opened 2019-10-01
Add ability to approve a solution based on tests

### LINKED ISSUE #4: ExerciseTest Module
At this point, the module (when used by another module) will implement the analyse function and compile the features found into discrete tests to run against the submission. It will also determine if it should disapprove a solution, but at this point it will not approve a solution. If it is not disapproved, it is referred. 

My thoughts for a roadmap for v1:

- [x] Convert the FastTest into an ExUnit test module to facilitate better testing
- [x] create a module attribute so that, if present, is able to automatically approve a submission if it passes a set of the feature tests (I think this would be technically simpler than another macro rule for ongoing maintenance and for people reading to understand the code)
- [x] tie this into the analyze pipeline, likely inserted around [this line](https://github.com/exercism/elixir-analyzer/blob/8bab9b444e7b52300b478b6450334d6413586276/lib/elixir_analyzer/exercise_test.ex#L131)

neenjaw:
This commit contains a few changes based on adding an ability of the analyzer to approve a solution.

- the DSL's feature attribute `severity` changed to `:info` from `:message` to reduce duplication of the word `message` since it was used to identify the unique markdown message to be displayed from `exercism/website-copy`
- better `@doc` written for elixir_analyzer.ex
- in `ElixirAnalyzer.ExerciseTest`, removed `failed_feature_test?/2` and expanded logic in `determine_status/3`
- in `ElixirAnalyzer.ExerciseTest.TwoFer` added an example how how to use the new feature
- ExUnit suite updated for new output

Close #4

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,7 +70,7 @@ This module contains macros for a DSL to be able to compare ideal solution featu
   feature "has spec" do
 
     mentor_message "elixir.two_fer.no_specification"
-    severity :message # or :disapprove or :refer
+    severity :info # or :disapprove or :refer
     match :all # :any, :none, :one
     # status :test # :skip
 
--- a/lib/elixir_analyzer.ex
+++ b/lib/elixir_analyzer.ex
@@ -1,6 +1,7 @@
 defmodule ElixirAnalyzer do
   @moduledoc """
-  Documentation for ElixirAnalyzer.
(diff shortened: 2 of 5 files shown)


neenjaw:
I'll leave this for another 24 hours and then I'll merge it if no suggestions for change.

PR #13, opened 2019-10-27
[Big PR] TwoFer Tests refined, ExerciseTest refactored/refined

neenjaw:
Pretext to changes:

I was able to obtain a large collection of two-fer submissions from @iHiD and subsequently ran the analyzer against them all, see [my test reesults repository](https://github.com/neenjaw/two-fer-test-data) (warning, it isn't really set up for much usable public consumption, but moreso as a record of the work).

Many changes were made in response to see how it fared against niche cases, erroring solutions, and odd constructs.

Working with the DSL, I also decided to change quite a few of the keywords to better suit the intent of the keyword so that it is more understandable how to interpret the test, and the result of the test.

---

Summary of the changes (from `git log`):

```
    Added some feature_tests:
    
    - added @spec feature
    - added bitstring guard
    - added default param, guard, wrong use of guard, aux func
    - added string interpolation and func header cases
```
```
    changes made to auto-approve, test clarified
```
```
    New DSL keywords added:
    
    - `suppress_if` allows tests to be silienced based on the status of other tests
```
```
   DSL changes:
    
    - `feature` is now `feature_test`
    - `:message` changed to `:comment` for congruity with exercism terminology
    - `:match` changed to `:find` for easier understanding the intent of the test
    - `:severity` changed to `:on_fail` to better represent
    
    Bugs Fixed:
    
    - suppressed results now filtered before they affect the submission determination
      - `filter_suppressed_results` created,
      - `append_test_comments`'s suppression code removed
```
```
    Guards simplified
    
    This test was first implemented when you couldn't suppress one test
    on another's result.  Now you can, so it makes it way easier to
    test for a specific pattern in the solution.
    
    So on the first pass, any appropriate guard is counted to pass the
    test "uses guards".  The second test "uses function level guard"
    then looks for the guard which is wanted in the optimal solution.
    
    Some README.md updates to refect current structure.
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -67,12 +67,12 @@ This module contains macros for a DSL to be able to compare ideal solution featu
 
   # This is DSL for describing the test to be done
   # This describes that a solution should have a typespec for the two_fer function
-  feature "has spec" do
+  feature_test "has spec" do
 
-    mentor_message "elixir.two_fer.no_specification"
-    severity :info # or :disapprove or :refer
-    match :all # :any, :none, :one
-    # status :test # :skip
+    status   :test # :skip -- optional
+    find     :all # :any, :none, :one
+    on_fail  :info # or :disapprove or :refer
+    comment  Constants.two_fer_no_specification # may also be a string
 
     # the form of the code that you are looking for
     # you may include more than one form block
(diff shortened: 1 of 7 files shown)


iHiD:
Whoop! I can't remember if Elixir's test runner is already in production? If not, shall we get it there? :)

neenjaw:
@iHiD Re: the test-runner, it is not running in production yet, I have a PR pending I re-pinged @NobbZ to it to get his input re: repo structure.

As far as this project goes, what is the process of testing it in the exercism framework? I think it should :crossed_fingers: run for two-fer solutions now.

iHiD:
There's an issue template on the automated-analysis repo you can use to open an issue then chat to @ccare on slack about it when he gets round to it (he's on holiday for a couple of days this week)

PR #20, opened 2020-09-04
Updating app to updated interface and docker requirements.

neenjaw:
A few things being updated here:
  - Dockerfile (@exercism/ops, please validate changes)
    - question: is a privileged user required? At the moment, i have a non-privileged user for the container user.
  - build/run bash scripts
  - elixir CLI update to take in output folder explicitly as third argument


neenjaw:
Sorry for lumping in workflow changes to this, but I realized that CI checks weren't being done on PR for a missed flag.  That is fixed now.

.github/workflows/elixir.yml
```diff
@@ -1,14 +1,14 @@
 name: Elixir CI
 
-on: push
+on: [push, pull_request]
```
ErikSchierboom, line 3: A slightly off-topic suggestion: maybe rename this workflow file to describe what it does? Like `tests.yml` or something like that?

ErikSchierboom (approved):

.github/workflows/elixir.yml
neenjaw: :+1:

PR #24, opened 2020-10-21
Unit test exercise tests

angelikatyborska:
This PR proves that it's possible to have unit tests for each exercise test module (e.g. `ElixirAnalyzer.ExerciseTest.TwoFer`) a predefined set of exercise submissions and expected analysis results.

Will it be enough for us to test the output of the whole analysis for an exercise, or would we want to find a solution that allows us to test single `feature`s?

If this is enough, then the next step would be to write some nice macros for generating those tests so that the current boilerplate in `ElixirAnalyzer.ExerciseTest.TwoFerTest` can be hidden away and doesn't need to be copied to each test file. Ideally it should also be possible to use `mix task` with the filepath and line number to only run one of the test cases.

One problem that I couldn't figure out is how to write a multiline string that contains another multiline string. I thought this would work:

```elixir
~S"""
  defmodule TwoFer do
    @doc \"\"\"
    Two-fer or 2-fer is short for two for one. One for you and one for me.
    \"\"\"
    @spec two_fer(binary()) :: binary()
    def two_fer(name \\ "you") when is_binary(name) do
      "One for #{name}, one for me"
    end
  end
"""
```

But that code causes `Code.string_to_quoted` to return
```
{:error, {2, "unexpected token: ", "\"\\\" (column 10, code point U+005C)"}}
```

### DIFF
--- a/lib/elixir_analyzer/exercise_test/two_fer.ex
+++ b/lib/elixir_analyzer/exercise_test/two_fer.ex
@@ -133,15 +133,16 @@ defmodule ElixirAnalyzer.ExerciseTest.TwoFer do
       defp _ignore(_ignore), do: _ignore
     end
 
-    form do
-      def two_fer(_ignore), do: _ignore
-      def _ignore(_ignore), do: _ignore
-    end
-
-    form do
-      def _ignore(_ignore), do: _ignore
-      def two_fer(_ignore), do: _ignore
-    end
+    #    # how to match functions named anything _but_ two_fer?
+    #    form do
+    #      def two_fer(_ignore), do: _ignore
+    #      def _ignore(_ignore), do: _ignore
+    #    end
+    #
+    #    form do
+    #      def _ignore(_ignore), do: _ignore
+    #      def two_fer(_ignore), do: _ignore
+    #    end
   end
 
   feature "uses string interpolation" do
--- a/lib/elixir_analyzer/exercise_test/two_fer.ex
+++ b/lib/elixir_analyzer/exercise_test/two_fer.ex
@@ -122,17 +122,6 @@ defmodule ElixirAnalyzer.ExerciseTest.TwoFer do
     form do
       defp _ignore(_ignore), do: _ignore
     end
-
-    #    # how to match functions named anything _but_ two_fer?
-    #    form do
-    #      def two_fer(_ignore), do: _ignore
-    #      def _ignore(_ignore), do: _ignore
-    #    end
-    #
-    #    form do
-    #      def _ignore(_ignore), do: _ignore
-    #      def two_fer(_ignore), do: _ignore
-    #    end
   end
 
   feature "uses string interpolation" do
--- a/lib/elixir_analyzer/exercise_test/two_fer.ex
+++ b/lib/elixir_analyzer/exercise_test/two_fer.ex
@@ -132,6 +132,16 @@ defmodule ElixirAnalyzer.ExerciseTest.TwoFer do
     form do
       defp _ignore(_ignore), do: _ignore
     end
+
+    form do
+      def two_fer(_ignore), do: _ignore
+      def _ignore(_ignore), do: _ignore
+    end
+
+    form do
+      def _ignore(_ignore), do: _ignore
+      def two_fer(_ignore), do: _ignore
+    end
   end
 
   feature "uses string interpolation" do
--- a/lib/elixir_analyzer/exercise_test/two_fer.ex
+++ b/lib/elixir_analyzer/exercise_test/two_fer.ex
@@ -48,12 +48,7 @@ defmodule ElixirAnalyzer.ExerciseTest.TwoFer do
       def two_fer(_ignore \\ "you")
     end
 
-    # function without a guard
-    form do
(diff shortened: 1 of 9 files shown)


lib/elixir_analyzer/exercise_test/two_fer.ex
```diff
@@ -132,6 +122,17 @@ defmodule ElixirAnalyzer.ExerciseTest.TwoFer do
...
+    #    # how to match functions named anything _but_ two_fer?
+    #    form do
+    #      def two_fer(_ignore), do: _ignore
+    #      def _ignore(_ignore), do: _ignore
+    #    end
+    #
+    #    form do
+    #      def _ignore(_ignore), do: _ignore
+    #      def two_fer(_ignore), do: _ignore
+    #    end
```
angelikatyborska, line 135: I wanted to extend the check about using auxiliary functions to also detect public functions that are not `two_fer/1`, but currently there is no way to match a public function that is called anything but `two_fer`. Do you think it would be possible to add such functionality? Or maybe this could be achieved already by writing two separate features and using `suppress_if` on one of them?

test/elixir_analyzer/exercise_test/two_fer_test.exs
```diff
@@ -0,0 +1,260 @@
...
+        #          end
+        #        end,
+        #        defmodule TwoFer do
+        #          def foo(_a) do
+        #          end
+        #
+        #          def two_fer(name \\ "you") do
+        #            "One for #{name}, one for me"
+        #          end
+        #        end
```
angelikatyborska, line 228: I would like for those test cases to also pass, but currently they fail due to: https://github.com/exercism/elixir-analyzer/pull/24/files#r512786752

test/elixir_analyzer/exercise_test/two_fer_test.exs
```diff
@@ -0,0 +1,260 @@
...
+  #        def two_fer(name \\ "you") do
+  #          if is_bitstring(name) do
+  #            "One for #{name}, one for me"
+  #          else
+  #            raise FunctionClauseError
+  #          end
+  #        end
+  #      end
+  #    end
+  #  end
```
angelikatyborska, line 259: This test currently fails and I think that might be a bug in the analyzer... it fails because the status of the analysis is `:refer` instead of `:disapprove`. I think the `:refer` status comes from the feature `"uses function level guard"`, but I expected `:disapprove` to always take priority over `:refer`. Since `:refer` means it's impossible to know if the solution is optimal or not, if two feature checks fail, one with `on_fail` of `:refer`, and one `:disapprove`, it _can_ be known for sure that the solution is not optimal.

angelikatyborska:
I created a `test_exercise_analysis` macro for writing test cases for exercise test modules. I used the first thing that came to my mind when it comes to the API of that macro, I'm open to suggestions for improvements.

I also added some documentation about writing those tests, and I left a few question open about details unrelated to the unit tests themselves, but related to the analyzer and the exercise test module for two-fer.

lib/elixir_analyzer/exercise_test/two_fer.ex
```diff
@@ -66,11 +61,6 @@ defmodule ElixirAnalyzer.ExerciseTest.TwoFer do
         _ignore
       end
     end
-
-    # function one-liner
-    form do
-      def two_fer(_ignore \\ "you") when _ignore, do: _ignore
-    end
```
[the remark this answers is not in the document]
neenjaw: It probably can, it would be a compile-time error since the `forms` from the test file are read at that time.  But could do some sort of check to see if the same form already exists.

lib/elixir_analyzer/exercise_test/two_fer.ex
neenjaw: I think that was one of my thoughts behind adding the `suppress_if` option, so that it would be easier to write tests like what you are describing

test/elixir_analyzer/exercise_test/two_fer_test.exs
neenjaw: Not sure if a bug, I think it was a how the spec was defined for V2.  I think you are right though, for V3, :disapprove should take precidence over :refer

neenjaw:
I can probably look at this in depth this afternoon/tonight.

neenjaw (approved):
This is a great addition, sorry for the wait

lib/elixir_analyzer/exercise_test/two_fer.ex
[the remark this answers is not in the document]
angelikatyborska: I created an issue for this: https://github.com/exercism/elixir-analyzer/issues/25

lib/elixir_analyzer/exercise_test/two_fer.ex
angelikatyborska: I came to the conclusion that it's impractical or even impossible to use `suppress_if` for this purpose. You would need to have a feature and a counter-feature for each possible combination of functions within the module. Also each feature needs to produce a comment, so it's not a good idea to add features only to be used for `suppress_if` of another feature.

PR #30, opened 2020-11-28
Add PacmanRules

angelikatyborska:
This is my attempt at implementing the following recommended analysis for the Pacman Rules (booleans) exercise:

https://github.com/exercism/v3/blob/6576904cbe3ddae26b23a418da8e012162980c4a/languages/elixir/exercises/concept/pacman-rules/.meta/design.md#L28-L32

### DIFF
--- a/test/elixir_analyzer/exercise_test/pacman_rules_test.exs
+++ b/test/elixir_analyzer/exercise_test/pacman_rules_test.exs
@@ -1,0 +1,104 @@
+defmodule ElixirAnalyzer.ExerciseTest.PacmanRulesTest do
+  use ElixirAnalyzer.ExerciseTestCase,
+    exercise_test_module: ElixirAnalyzer.ExerciseTest.PacmanRules
+
+  test_exercise_analysis "example solution",
+    comments: [],
+    status: :approve do
+    defmodule Rules do
+      def eat_ghost?(power_pellet_active, touching_ghost) do
+        power_pellet_active and touching_ghost
+      end
+
+      def score?(touching_power_pellet, touching_dot) do
+        touching_power_pellet or touching_dot
+      end
+
+      def lose?(power_pellet_active, touching_ghost) do
+        not power_pellet_active and touching_ghost
+      end
+
+      def win?(has_eaten_all_dots, power_pellet_active, touching_ghost) do
+        has_eaten_all_dots and not lose?(power_pellet_active, touching_ghost)
+      end
+    end
+  end
+
+  describe "requires strictly boolean operators" do
+    test_exercise_analysis "detects &&",
+      comments_include: [Constants.pacman_rules_use_strictly_boolean_operators()] do
+      [
+        defmodule Rules do
+          def eat_ghost?(power_pellet_active, touching_ghost) do
+            power_pellet_active && touching_ghost
+          end
+        end,
+        defmodule Rules do
+          def eat_ghost?(power_pellet_active, touching_ghost) do
+            power_pellet_active and touching_ghost
+          end
+
+          def lose?(power_pellet_active, touching_ghost) do
+            not power_pellet_active && touching_ghost
+          end
+        end,
+        defmodule Rules do
+          def eat_ghost?(power_pellet_active, touching_ghost) do
+            Kernel.&&(power_pellet_active, touching_ghost)
+          end
+        end
+      ]
+    end
+
+    test_exercise_analysis "detects ||",
+      comments_include: [Constants.pacman_rules_use_strictly_boolean_operators()] do
+      [
+        defmodule Rules do
+          def score?(touching_power_pellet, touching_dot) do
+            touching_power_pellet || touching_dot
+          end
+        end,
+        defmodule Rules do
+          def score?(touching_power_pellet, touching_dot) do
+            Kernel.||(touching_power_pellet, touching_dot)
+          end
+        end
+      ]
+    end
+
+    test_exercise_analysis "detects !",
+      comments_include: [Constants.pacman_rules_use_strictly_boolean_operators()] do
+      [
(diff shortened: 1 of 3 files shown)


test/elixir_analyzer/exercise_test/pacman_rules_test.exs
```diff
@@ -0,0 +1,104 @@
...
+          def eat_ghost?(power_pellet_active, touching_ghost) do
+            !!power_pellet_active and !!touching_ghost
+          end
+        end,
+        defmodule Rules do
+          def lose?(power_pellet_active, touching_ghost) do
+            Kernel.!(power_pellet_active) and touching_ghost
+          end
+        end
+      ]
```
angelikatyborska, line 101: The first three test cases fail, while the last two work and I have no idea why. I would think I cannot go wrong with such a simple form:
```elixir
form do
  !_ignore
end
```

But apparently I can...

```
  1) test requires strictly boolean operators detects ! 1 (ElixirAnalyzer.ExerciseTest.PacmanRulesTest)
     test/elixir_analyzer/exercise_test/pacman_rules_test.exs:72
     Assertion with in failed
     code:  assert comment in result.comments
     left:  "elixir.pacman_rules.use_strictly_boolean_operators"
     right: []
     stacktrace:
       (elixir 1.10.4) lib/enum.ex:783: Enum."-each/2-lists^foreach/1-0-"/2
       (elixir 1.10.4) lib/enum.ex:1396: Enum."-map/2-lists^map/1-0-"/2
       test/elixir_analyzer/exercise_test/pacman_rules_test.exs:72: (test)

  * test requires strictly boolean operators detects ! 2 (0.1ms)

  2) test requires strictly boolean operators detects ! 2 (ElixirAnalyzer.ExerciseTest.PacmanRulesTest)
     test/elixir_analyzer/exercise_test/pacman_rules_test.exs:77
     Assertion with in failed
     code:  assert comment in result.comments
     left:  "elixir.pacman_rules.use_strictly_boolean_operators"
     right: []
     stacktrace:
       (elixir 1.10.4) lib/enum.ex:783: Enum."-each/2-lists^foreach/1-0-"/2
       (elixir 1.10.4) lib/enum.ex:1396: Enum."-map/2-lists^map/1-0-"/2
       test/elixir_analyzer/exercise_test/pacman_rules_test.exs:77: (test)

  * test requires strictly boolean operators detects ! 3 (0.1ms)

  3) test requires strictly boolean operators detects ! 3 (ElixirAnalyzer.ExerciseTest.PacmanRulesTest)
     test/elixir_analyzer/exercise_test/pacman_rules_test.exs:82
     Assertion with in failed
     code:  assert comment in result.comments
     left:  "elixir.pacman_rules.use_strictly_boolean_operators"
     right: []
     stacktrace:
       (elixir 1.10.4) lib/enum.ex:783: Enum."-each/2-lists^foreach/1-0-"/2
       (elixir 1.10.4) lib/enum.ex:1396: Enum."-map/2-lists^map/1-0-"/2
       test/elixir_analyzer/exercise_test/pacman_rules_test.exs:82: (test)
```

neenjaw: Interesting behavior.  Off the top of my head I don't know either.

AST of the form:
```elixir
{:__block__, [], [{:!, [line: 1], [{:_ignore, [line: 1], nil}]}]}

# Which should get reduced by the analyzer to:
{:!, _, [_]}
```

AST of the module: 
```elixir
{:defmodule, [line: 1],
 [
   {:__aliases__, [line: 1], [:Rules]},
   [
     do: {:def, [line: 2],
      [
        {:lose?, [line: 2],
         [
           {:power_pellet_active, [line: 2], nil},
           {:touching_ghost, [line: 2], nil}
         ]},
        [
          do: {:and, [line: 3],
           [
             {:!, [line: 3], [{:power_pellet_active, [line: 3], nil}]},   # Which then should match here
             {:touching_ghost, [line: 3], nil}
           ]}
        ]
      ]}
   ]
 ]}
```

I would suspect that somewhere when the `form` is compiled, it isn't forming the correct pattern.  Maybe I didn't account for unary functions with the pattern matching? Maybe it is making the wrong pattern to match against?

neenjaw: Putting an inspect [here](https://github.com/exercism/elixir-analyzer/blob/ce19fd3e64039ca5aafaee41e3244b6390e80aa0/lib/elixir_analyzer/exercise_test.ex#L99) would let you get the outputs of each form **when you compile**.

The output might be a bit noisy, so it might be worthwhile to comment out other cases and temporarily remove the other modules that might get compiled to isolate the pattern generated by the form.

angelikatyborska: So you're saying that 
```
{:__block__, [], [{:!, [line: 1], [{:_ignore, [line: 1], nil}]}]}
```
should get reduced to 
```
{:!, _, [_]}
```

But in reality, it gets reduced to:
```
[{:!, _, [_]}]
```

So it can only ever match if it's used on its own in a single line of code. This is unlike regular function calls, which aren't wrapped in a list and thus behave like expected:
```
form do
  foo(_ignore)
end

# gets reduced to
{:foo, _, [_]}
```

I'm not sure at the moment how to work around that...

neenjaw: You can't, it is a missed case in my design. It should have resulted as I said. So now we know where the bug is. 👍

neenjaw: Yay, bug resolved! :boot: :beetle:

neenjaw (approved):

PR #31, opened 2020-11-29
fix AST pattern generation for some block expressions

neenjaw:
when creating AST, elixir handles some expressions differently from others depending on what it determines the contend of a `do...end` block to contain.  For example:
```elixir
foo do
  !boolean
end
```
results in :
```elixir
{:foo, [],
 [
   [
     do: {:__block__, [],
      [{:!, [context: Elixir, import: Kernel], [{:boolean, [], Elixir}]}]}
   ]
 ]}
```
Whereas:
```elixir
foo do
  bar(boolean)
end
```
Results in:
```elixir
{:foo, [], [[do: {:bar, [], [{:boolean, [], Elixir}]}]]}
```

This caused a pattern match problem when applied to the `form` block used by the analyzer.  Some unary functions (like `!`) wer not being unwrapped from the block causing them never to match. This patch fixes by unwrapping `form` blocks with only 1 expression even if it was determined by Elixir to wrap it in a block.

---

@angelikatyborska I did add in your pacman test to this locally and it did seem to pass all of the tests that you had fail, but if you test this patch it may be worthwhile to temporarily add in your analyzer module / test file to just make sure. I haven't missed something else.

### DIFF
--- a/lib/elixir_analyzer/exercise_test.ex
+++ b/lib/elixir_analyzer/exercise_test.ex
@@ -76,8 +76,17 @@ defmodule ElixirAnalyzer.ExerciseTest do
 
     {ast, block_params} =
       case ast do
-        {:__block__, _, params} -> {params, length(params)}
-        _ -> {ast, false}
+        {:__block__, _, params} ->
+          case length(params) do
+            1 ->
+              {hd(params), false}
+
+            len ->
+              {params, len}
+          end
+
+        _ ->
+          {ast, false}
       end
 
     find_ast_string =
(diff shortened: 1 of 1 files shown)


lib/elixir_analyzer/exercise_test.ex
```diff
@@ -76,8 +76,17 @@ defmodule ElixirAnalyzer.ExerciseTest do
...
+          case length(params) do
+            1 ->
+              {hd(params), false}
+
+            len ->
+              {params, len}
+          end
+
+        _ ->
+          {ast, false}
```
angelikatyborska, line 89: It makes my tests work and doesn't break anything else as far as I can tell. I think we will have to write analyzers for a few more exercises to really test it.

Style-wise, I would propose to remove the nested case and do pattern-matching instead:
```suggestion
        {:__block__, _, [param]} ->
          {param, false}

        {:__block__, _, [_ | _] = params}  ->
          {params, length(params)}

        _ ->
          {ast, false}
```

angelikatyborska (approved):

lib/elixir_analyzer/exercise_test.ex
neenjaw: ya, I just wanted to make sure that it actually worked before worrying too much about aesthetic

neenjaw: I like the suggestion tho

PR #32, opened 2020-11-29
Adding the ensure_call test

neenjaw:
So far:
- refactoring to slim the `exercise_test.ex` into a leaner file.
- Added stubs for the `ensure_call` macro
- added @angelikatyborska's pacman test with feature fix applied

---

@angelikatyborska I don't know if you want to be subscribed to this, but I thought I would open this draft PR so you could see how it's going.


config/exercise_data.json
```diff
@@ -1,6 +1,6 @@
...
-        "code_file": "two_fer.ex",
-        "analyzer_module": "ElixirAnalyzer.ExerciseTest.TwoFer"
-    }
-}
\ No newline at end of file
+  "two-fer": {
+    "code_file": "two_fer.ex",
+    "analyzer_module": "ElixirAnalyzer.TestSuite.TwoFer"
+  }
+}
```
angelikatyborska, line 6: Oh, I didn't notice this file before. Could you add the `pacman-rules` exercise here? Normally I would do it on my own, on my own branch, but since you're changing the module names, it makes more sense to add it on this branch.

lib/test_suite/pacman_rules.ex
```diff
@@ -1,4 +1,4 @@
-defmodule ElixirAnalyzer.ExerciseTest.PacmanRules do
+defmodule ElixirAnalyzer.TestSuite.PacmanRules do
```
angelikatyborska, line 1: The examples in `docs` will need to be updated with the same module name change

angelikatyborska (commented):
Thanks! I appreciate being subscribed to this PR. I am looking forward to this new macro so that I can write more exercise analyzers for v3.

PR #37, opened 2021-01-30
New location of analyzer comments

angelikatyborska:
Analyzer comments were moved in that PR: https://github.com/exercism/website-copy/pull/1898

On top of fixing that location, this PR:
- makes the external (analyzer comments) test run on CI so we don't forget that some comments are missing
- fixes the paths of some existing comments (exercise names use dashes instead of underscores)
- fixes an unrelated unit test failure about extracting the line number

### DIFF
--- a/.github/workflows/elixir_test.yml
+++ b/.github/workflows/elixir_test.yml
@@ -18,4 +18,4 @@ jobs:
         mix local.hex --force
         mix deps.get
     - name: Run Tests
-      run: mix test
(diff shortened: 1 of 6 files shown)


neenjaw (approved):
Can we duplicated the test workflow so that we have one that tests the external and one that doesn't?

angelikatyborska:
So basically one with `mix test --exclude external` and one with `mix test --only external`? Yeah, sure.

PR #44, opened 2021-02-27
upgrade to newest spec

neenjaw:
status removed
on_fail removed
type added

submission simplified
elixir_analyzer simplified flow
exercise test now adds comments based on test pass/fail
rules updates


neenjaw:
Things still to do:
- update docs
- update tests
- add smoke test

docs/step-03/step-03.md
```diff
@@ -228,15 +228,16 @@ defmodule ElixirAnalyzer.TestSuite.Example do
...
-##### `on_fail`
+##### `type`
 
-`on_fail` tells our analyzer how to treat the solution when it fails the test. There are 3 accepted actions:
+`type` tells our analyzer how to the solution the purpose of the test. There are 3 accepted actions:
 
-- `:info` - append the comment associated with this test
-- `:disapprove` - disapprove the submission if the test fails, append the comment associated with this test
-- `:refer` - the analyzer is unable to determine if the submission should be approved or disapproved, flag that it requires manual attention.
+- `:celebratory` - if the test passes, append a comment to celebrate the student's acheivement
```
angelikatyborska, line 235:
```suggestion
- `:celebratory` - if the test passes, append a comment to celebrate the student's achievement
```

lib/elixir_analyzer/submission.ex
```diff
@@ -8,130 +8,129 @@ defmodule ElixirAnalyzer.Submission do
...
+  # Essential
+  defp build_summary_response(t = %{essential: 0}, [:essential | rest], acc),
+    do: build_summary_response(t, rest, acc)
 
-      true ->
-        do_append_comment(r, %{"comment" => comment, "params" => params})
-    end
-  end
+  defp build_summary_response(_, [:essential | _], acc),
+    do: "#{acc} Check out the comments for things to fix." |> String.trim_leading()
```
angelikatyborska, line 121: I see that for essential, actionable, and informative comments, the accumulator is used when producing the summary, but there is no recursive call. For celebratory comments, the accumulator is set to "🎉" and there is a recursive call. Does that mean this is the expected output?

0 celebratory, 1 essential, 0 actionable:
```
"Check out the comments for things to fix."
```

0 celebratory, 1 essential, 1 actionable:
```
"Check out the comments for things to fix."
```

1 celebratory, 1 essential, 1 actionable:
```
"🎉 Check out the comments for things to fix."
```

The last case seems a bit inappropriate.

I think it would be nice to have unit tests that will demonstrate how choosing the summary works with different combinations of different comment types. I agree with the current behavior of the summary when there are essential, actionable, and/or informative comments, but I think for celebratory comments, we should only have a celebratory summary when there are no other types of comments.

lib/elixir_analyzer/submission.ex
```diff
@@ -8,130 +8,129 @@ defmodule ElixirAnalyzer.Submission do
...
 
-  def set_analyzed(r = %__MODULE__{}, value) when is_boolean(value) do
-    %{r | analyzed: value}
-  end
+  # Informative
+  defp build_summary_response(t = %{informative: 0}, [:informative | rest], acc),
+    do: build_summary_response(t, rest, acc)
+
+  defp build_summary_response(_, [:informative | _], acc),
+    do: "#{acc} Check out the comments for some things to learn." |> String.trim_leading()
```
angelikatyborska, line 135: I was wondering if maybe those summaries should be fetched from https://github.com/exercism/website-copy/tree/main/analyzer-comments/elixir in the same way as comments are. What do you think?

lib/elixir_analyzer/submission.ex
```diff
@@ -8,130 +8,129 @@ defmodule ElixirAnalyzer.Submission do
...
-    |> Map.keys()
-    |> Enum.all?(fn key -> Kernel.is_atom(key) or Kernel.is_binary(key) end)
-    |> case do
-      false ->
-        raise ArgumentError, "key type must be atom or binary"
-
-      true ->
-        do_prepend_comment(r, %{"comment" => comment, "params" => params})
+    case summary do
+      nil -> "Submission analyzed. No automated suggestions. Great work! 🚀"
```
angelikatyborska, line 101: I'm not sure if we should say here "great work!". It could be that somebody submitted such a weird solution that we didn't foresee it in the analyzer :D. Maybe we could be neutral?
```suggestion
      nil -> "Submission analyzed. No automated suggestions found."
```

docs/step-03/step-03.md
```diff
@@ -228,15 +228,16 @@ defmodule ElixirAnalyzer.TestSuite.Example do
...
+- `:actionable` - append a comment with an actionable improvement to the solution
+- `:informative` - append a comment with a point of learning
+- `:essential` - append a comment which soft-blocks the student on the website
```
angelikatyborska, line 238: I think it would be useful to be explicit about how the behavior of celebratory comments differs from the rest:
```suggestion
- `:actionable` - if the test fails, append a comment with an actionable improvement to the solution
- `:informative` - if the test fails, append a comment with a point of learning
- `:essential` - if the test fails, append a comment which soft-blocks the student on the website
```

lib/elixir_analyzer.ex
```diff
@@ -49,25 +51,32 @@ defmodule ElixirAnalyzer do
...
   @spec analyze_exercise(String.t(), String.t(), String.t(), keyword()) :: Submission.t()
   def analyze_exercise(exercise, input_path, output_path, opts \\ []) do
+    Logger.info("Analyzing #{exercise}")
+
+    Logger.debug("Input: #{input_path}, Output: #{output_path}")
+
     params = get_params(exercise, input_path, output_path, opts)
 
-    s =
+    submission =
```
angelikatyborska, line 60: I see you lived to regret choosing single-letter variable names 😁

angelikatyborska (commented):
I'm loving the improved logging 🎉

lib/elixir_analyzer.ex
neenjaw: I would like to think all of those functions started much smaller at one time, then they got bigger and I never made the change. 🤷🏻 In reality, it was probably just a bad choice at the time.

lib/elixir_analyzer/submission.ex
neenjaw: I would be up for that if it were possible.  it would be better to decouple them.

@iHiD Can this be a thing? Is it a thing already?

lib/elixir_analyzer/submission.ex
neenjaw: Ya, I was just trying to give it some sense of personalized message based on the results of the analyzer. The last case doesn't feel right to me either. Will iterate

neenjaw:
I simplified the summary.  It will return on the first type matching when there are comments.

angelikatyborska:
👍 to the summary change.

I just noticed that there's a compilation warning:

```
warning: Logger.warning/1 is undefined or private. Did you mean one of:

      * warn/1
      * warn/2

Found at 4 locations:
  lib/elixir_analyzer.ex:133: ElixirAnalyzer.init/1
  lib/elixir_analyzer.ex:166: ElixirAnalyzer.check/2
  lib/elixir_analyzer.ex:179: ElixirAnalyzer.check/2
  lib/elixir_analyzer.ex:196: ElixirAnalyzer.analyze/2
```

angelikatyborska (approved):

lib/elixir_analyzer/submission.ex
iHiD: Not currently. The idea is that the summaries will potentially be personalised to the situation, rather than "predicable" things. But maybe further down the line once we've got some examples of how tracks are using them.

PR #46, opened 2021-03-02
Update exercise config

neenjaw:
Small config change


neenjaw:
is take_a_number's correct?

angelikatyborska:
> is take_a_number's correct?

Yes.

Now that we have this information per-exercise in `config.json` in the track, it would be nice if we found a way to use that instead repeating the information in the analyzer's config 🤔.

angelikatyborska (approved):

PR #47, opened 2021-03-06
Use solution config to find code file

neenjaw:
Since we are already tracking this file thought the config, use that.

### DIFF
--- a/lib/elixir_analyzer.ex
+++ b/lib/elixir_analyzer.ex
@@ -14,7 +14,7 @@ defmodule ElixirAnalyzer do
   # defaults
   @exercise_config Application.get_env(:elixir_analyzer, :exercise_config)
   @output_file "analysis.json"
-  @lib_dir "lib"
+  @meta_config ".meta/config.json"
 
   @doc """
   This is the main entry point to the analyzer.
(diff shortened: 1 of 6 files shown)


neenjaw:
@iHiD, I had heard that solutions get the .meta/config when being analyzed.  Is that true?

iHiD:
It is!

lib/elixir_analyzer.ex
```diff
@@ -129,24 +129,41 @@ defmodule ElixirAnalyzer do
...
 
   # When file is nil, pull code params from config file
   defp do_init(%{file: nil} = params, exercise_config) do
-    {
-      Path.join(params.path, @lib_dir),
-      exercise_config.code_file,
-      exercise_config.analyzer_module
-    }
+    meta_config = Path.join(params.path, @meta_config) |> File.read!() |> Jason.decode!()
+    relative_code_path = meta_config["files"]["solution"] |> hd()
```
angelikatyborska, line 160: We do have one or two exercises that have two solution files. Does this mean that those exercises cannot be analyzed?

neenjaw: Hmmm..

iHiD: (The solutions files key is an array, as I'm sure you're both aware)

You could always concat two files together and write out a final one, if Elixir supports that sort of approach.

neenjaw: I can't recall them off the top of my head, but do those exercises have two files that the student must write in or just one? If just one, we only need to analyze the file that the student adds their code to. If there are two files then maybe we can concat them like the representer

angelikatyborska: One of them is `dot-dsl` which provides two files, and looking at community solutions, my own old solution included, people modify code in both files, e.g.: https://exercism.io/tracks/elixir/exercises/dot-dsl/solutions/3ffa33ffb0bc4c35b4dc5dda12df64cb

angelikatyborska: The other one is `zipper` and there people only write code in one of the files. That's that, just those two exercises.

neenjaw: So, three options off the top of my head:
1) Do nothing for now, they will fail silently and when we want to implement analyzer features for them, then deal with it (easiest -- my vote at this moment)
2) Concatenate them then analyze, I'd need to review how `Code.string_to_quoted` works in that situation to see if breaks the analzyer
3) Merge them into a single file, making them sub-modules of the main module (eh, this makes these big files even bigger/complicated) when they are already reasonably hard exercises

angelikatyborska: Since this is an existing limitation of the analyzer, not introduced in this PR, I'm ok with option 1.

angelikatyborska (approved):

PR #49, opened 2021-03-07
New analyzer function: `assert_call`

neenjaw:
Introduces a new analyzer function, `assert_call`:

This PR proposes a new function which allows an analyzer test suite to assert that a function is called.  It may be a public or private function. It may also be a function call with the same module or another globally available module function.

Adds some basic tests.  Probably needs further testing, test suite expansion.


lib/test_suite/analyzer_verification/assert_call.ex
```diff
@@ -0,0 +1,34 @@
+defmodule ElixirAnalyzer.TestSuite.AnalyzerVerification.AssertCall do
+  @dialyzer generated: true
+  @moduledoc """
+  This is an exercise analyzer extension module for the concept exercise Pacman Rules
```
angelikatyborska, line 4: I guess this comment is not correct?

lib/test_suite/analyzer_verification/assert_call.ex
```diff
@@ -0,0 +1,34 @@
...
+    comment "didn't find a local call to helper/0 within function/0"
+  end
+
+  assert_call "finds call to private local helper function within function" do
+    type :informational
+    calling_fn :global, &AssertCallVerification.function/0
+    called_fn :local, &AssertCallVerification.private_helper/0
+    comment "didn't find a local call to private_helper/0 within function"
+  end
+end
```
angelikatyborska, line 34: The purpose of this module is to serve as a basis for the test `ElixirAnalyzer.ExerciseTest.AssertCallTest`, right? It's not actually part of the source code of the analyzer. If that's the case, I think it belongs in `test/support` instead.

lib/test_suite/analyzer_verification/assert_call.ex
```diff
@@ -0,0 +1,34 @@
...
+  assert_call "finds call to private local helper function" do
+    type :informational
+    called_fn :local, &AssertCallVerification.private_helper/0
+    comment "didn't find a local call to private_helper/0"
+  end
+
+  assert_call "finds call to local helper function within function" do
+    type :informational
+    calling_fn :global, &AssertCallVerification.function/0
+    called_fn :local, &AssertCallVerification.helper/0
```
angelikatyborska, line 24: What's the difference between `:global` and `:local`? are both options valid for both `calling_fn` and `called_fn`?

lib/test_suite/analyzer_verification/assert_call.ex
```diff
@@ -0,0 +1,34 @@
+defmodule ElixirAnalyzer.TestSuite.AnalyzerVerification.AssertCall do
+  @dialyzer generated: true
+  @moduledoc """
+  This is an exercise analyzer extension module for the concept exercise Pacman Rules
+  """
+
+  use ElixirAnalyzer.ExerciseTest
+
+  assert_call "finds call to local helper function" do
```
angelikatyborska, line 9: Would it be possible to add something that asserts a function wasn't called anywhere? This would be crucial for writing analyzers for exercises like the concept exercise for recursion or `list-ops`, to verify that functions from `Enum` aren't used.

lib/test_suite/analyzer_verification/assert_call.ex
neenjaw: Ya, I don't know if I'm 100% happy with this part of the DSL yet. Basically I need a way to specify a the function. So if it is local, it strips off the module name to be matched. 

Maybe it would be better as a keyword list where the programmer specifies the function. I don't know. 

So while I think `:global` and `:local` would work for both directives right now, I'm not sure it makes sense for the calling_fn.

🤷‍♂️

lib/test_suite/analyzer_verification/assert_call.ex
neenjaw: Yea I have something in the underlying code to do that, not sure if I exposed it yet.

neenjaw: I've exposed this as `assert_no_call` for now

lib/test_suite/analyzer_verification/assert_call.ex
neenjaw: I've simplified this.  See new.

lib/elixir_analyzer/exercise_test/assert_call.ex
```diff
@@ -0,0 +1,137 @@
...
+    comment: "message"
+  }
+  """
+  defmacro assert_call(description, do: block) do
+    parse(description, block, true)
+  end
+
+  defmacro assert_no_call(description, do: block) do
+    parse(description, block, false)
+  end
```
neenjaw, line 34: macros to assert a function is called and that a function is not called.

test/support/analyzer_verification/assert_call.ex
```diff
@@ -0,0 +1,47 @@
...
+    called_fn name: :private_helper
+    comment "didn't find a local call to private_helper/0"
+  end
+
+  assert_call "finds call to local helper function within function" do
+    type :informational
+    calling_fn module: AssertCallVerification, name: :function
+    called_fn name: :helper
+    comment "didn't find a local call to helper/0 within function/0"
+  end
```
neenjaw, line 26: updated DSL with simpler function specification

neenjaw:
@angelikatyborska I think this is ready for a look.  I have tests covering I think the major use cases of this feature.  If I've missed something we can add it in as we go along.

At this point I decided to abandon testing for specific arguments and function arities.  Those would maybe be nice-to-haves in the future.

Let me know what you think.

test/support/analyzer_verification/assert_not_call.ex
```diff
@@ -0,0 +1,34 @@
+defmodule ElixirAnalyzer.Support.AnalyzerVerification.AssertNoCall do
```
angelikatyborska, line 1: Nitpick: filename doesn't match the module name, `_not_` -> `_no_`, same applies to the file `test/elixir_analyzer/exercise_test/assert_not_call_test.exs`

angelikatyborska (approved):
I reviewed the tests to get a feel for the DSL. I like it, I think this is exactly what we need to write the most important analyses for exercises. I played around with the tests for a sanity check, but I didn't review all the code changes because it would take me too long to understand it all for that to be productive

> At this point I decided to abandon testing for specific arguments and function arities. 

Agreed!

Let's merge this, try to use it in real exercises and see what happens! Looks exciting :)

PR #51, opened 2021-04-06
add wildcard function matcher

neenjaw:
When specifying a function to be called or not called you can now use `:*` to specify a wilcard function name.  This will then match any function from a specified module;

```elixir
  assert_no_call "does not call any function from List module" do
    type :informational
    called_fn module: List, name: :*
    comment "don't call List module functions"
  end
```

### DIFF
--- a/lib/elixir_analyzer/exercise_test/assert_call/compiler.ex
+++ b/lib/elixir_analyzer/exercise_test/assert_call/compiler.ex
@@ -130,6 +130,13 @@ defmodule ElixirAnalyzer.ExerciseTest.AssertCall.Compiler do
   def matching_function_call?(_node, nil), do: false
 
   def matching_function_call?(
+        {{:., _, [{:__aliases__, _, module_path}, _name]}, _, _args},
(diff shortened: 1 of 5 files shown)


angelikatyborska (approved):
I let myself add an extra test for `assert_call` with the wildard. It works 👍

This will unfortunately prevent us from using `assert_call` with functions that are actually called `*`, but I can't think of any other wildcard symbol that wouldn't have this problem...

neenjaw:
Yea, I thoguht of that, but the only one I know of is `Kernel.*/2` and I hope we don't need to write a rule for that.  We could use a atom enclosed in quotes too, `:"*any"` or some such if we wanted to be extra careful.

angelikatyborska:
Maybe `name: :_ignore` could serve as a wildcard? Like in the `form` expressions in the `feature` macro. But I'm also fine with keeping `:*`

neenjaw:
What about just `:_`?

angelikatyborska:
I like that!

PR #52, opened 2021-04-07
Should support functions in assert_call test cases now

neenjaw:


neenjaw:
I think this addition should let you use the function constants

neenjaw:
Should work for realz now

PR #54, opened 2021-04-15
`assert_call` bugs

angelikatyborska:
A bug in the implementation of `assert_(no_)call`. For example, `called_fn module: Enum, name: :map` is going to match the usage of any function named `map`, even if it has nothing to do with the `Enum` module. My expectation is that it would only match calls to `Enum.map`, and allow using any other `map` functions even if somebody tries to cheat by aliasing or importing `Enum`. I would expect that to detect that, we would write a separate check that forbids aliases and imports of that module.

This bug also causes the usage of an underscore `_`, e.g. in `[h | _] = list` to trigger checks for `called_fn module: Enum, name: :_`.

@neenjaw is this something you could look into, since you're already familiar with that code? I'm creating this PR with failing tests, feel free to use it and push commits here.

### DIFF
--- a/test/support/constants.ex
+++ b/test/support/constants.ex
@@ -1,3 +1,5 @@
 defmodule ElixirAnalyzer.Support.Constants do
+  @moduledoc false
+
   def mock_constant(), do: "mock.constant"
 end


neenjaw:
I can look into it, yes. Probably not until the weekend at the earliest

I disagree with putting too much effort into dissuading people cheating with aliases because the student could use dynamic dispatch in any number of ways to call a function.

angelikatyborska:
I think I found a fix. When finding a call for a function when the call doesn't use the module name, we need to pattern-match that the given module is `nil` (so we're looking for a local call).

> I disagree with putting too much effort into dissuading people cheating with aliases because the student could use dynamic dispatch in any number of ways to call a function.

I promise to only put a tiny bit of effort into it 🤞. I won't touch the topic of dynamic dispatch in any way.

neenjaw:
I think this was closed automatically since the changes are already present on `main`

PR #56, opened 2021-04-23
Add Credo

angelikatyborska:
This PR adds credo and fixes all problems reported by it. It also adds the credo check to the CI.

I'm also sneaking in `.tool-versions`, the same as the test runner and representer have.


.credo.exs
```diff
@@ -0,0 +1,199 @@
...
+        excluded: [
+          ~r"/_build/",
+          ~r"/deps/",
+          ~r"/node_modules/",
+          # exclude exercise analysis because of snippets in `feature do form do...`
+          # they can be unusual, e.g. `_ignore || _ignore`
+          "lib/elixir_analyzer/test_suite/",
+          # exclude exercise analysis test because of example solutions used in tests
+          # those mimic student solutions and might be "bad code" on purpose
+          "test/elixir_analyzer/test_suite/"
```
angelikatyborska, line 43: This is the only change in this config compared to the default config.

lib/elixir_analyzer/exercise_test.ex
```diff
@@ -23,6 +23,7 @@ defmodule ElixirAnalyzer.ExerciseTest do
   #
 
   defmacro __before_compile__(env) do
+    # credo:disable-for-previous-line Credo.Check.Refactor.CyclomaticComplexity
```
angelikatyborska, line 26: This is the only place where I was afraid of splitting this code to reduce cyclomatic complexity, so I took the easy way out...

.tool-versions
```diff
@@ -0,0 +1,2 @@
+elixir 1.11.3-otp-23
+erlang 23.2.6
```
neenjaw, line 2: 👍🏻 Looks good!

neenjaw (approved):
Looks good, forgot how opinionated credo is about `cond` vs `if`, but I'm good with it

PR #57, opened 2021-04-23
Ensure comments are unique

### LINKED ISSUE #55: Ensure comments are unique
Sometimes, a single check must be implemented with a combination of `assert_(no_)call` and `feature` calls. For example, checking that the `Enum` module is not used:
```elixir
  assert_no_call "does not call any Enum functions" do
    type :essential
    called_fn module: Enum, name: :_
    comment ElixirAnalyzer.Constants.bird_count_use_recursion()
  end

  feature "does not alias or import Enum" do
    find :none
    type :essential
    comment ElixirAnalyzer.Constants.bird_count_use_recursion()

    form do
      alias Enum, as: _ignore
    end
  end
```

Both checks return the same comment when failed. In this case, if somebody imported the `Enum` module but then also used it forgetting about the import, the comment would appear to that user twice.

To prevent the user from seeing the same comment twice, we should make sure that all values in the returned `comments` list are unique.

angelikatyborska:
Closes #55

### DIFF
--- a/lib/elixir_analyzer/submission.ex
+++ b/lib/elixir_analyzer/submission.ex
@@ -75,7 +75,14 @@ defmodule ElixirAnalyzer.Submission do
       end)
       |> Enum.into(%{})
 
-    %{submission | comments: submission.comments ++ [comment]}
+    comments =
+      if Enum.member?(submission.comments, comment) do
+        submission.comments
+      else
+        submission.comments ++ [comment]
+      end
+
+    %{submission | comments: comments}
   end
 
   defp get_summary(submission = %__MODULE__{halted: true, comments: comments})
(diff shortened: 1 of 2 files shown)


neenjaw:
Does the action type `:essential, :celebratory`...etc then belong to the comment rather than the test? MIght be worthwhile to think about refactoring it at somepoint.

neenjaw (approved):

angelikatyborska:
Yes, the type `:essential`, `:celebratory` etc. belongs to the comment (https://github.com/exercism/elixir-analyzer/blob/324c52eac2b444e3622d927e7f8cec9a5b2bd075/lib/elixir_analyzer/submission.ex#L73) but I think this is actually according to the specification. Here https://github.com/exercism/docs/blob/main/building/tooling/analyzers/interface.md#comments it says that `type` is a field on an object that's in the `comments` list.

PR #58, opened 2021-04-23
Add dialyzer and make it green

angelikatyborska:
I noticed that the instructions on how to write an analyzer extension mention that each one needs `@dialyzer generated: true`. I don't really understand why. Dialyzer is not used in this project. But if it's really needed, then maybe it will work to move this module attribute into `ElixirAnalyzer.ExerciseTest`s `__using__` so that we don't need to add it to every analyzer extension?

```elixir
  defmacro __using__(_opts) do
    quote do
      # other stuff...
      @dialyzer generated: true
    end
  end
```

However, because I am not sure _why_ this module attribute is needed, I wasn't able to check if after this change it will still serve its purpose, or maybe this breaks something.


neenjaw:
It's in there because ElixirLS used to continuously throw an error about the modules because when compiles it can produce a case statements that will always match a certain option. 

I think this didn't work before, but if it does now then 🎉

I'll test it out this weekend

angelikatyborska:
I see! I didn't know that ElixirLS runs dialyzer for all projects.

After some googling, I can't find any proof that `@dialyzer generate: true` is supposed to do anything. The only mentions of that snippet of code that GitHub finds are in this repository: https://github.com/search?p=1&q=%22%40dialyzer+generated%3A+true%22&type=Code

When I add dialyzer and run it against the main branch (where each module defines `@dialyzer generate: true` on its first line), I get this error:

```
:dialyzer.run error: Analysis failed with error:
Could not scan the following file(s):
  Bad function true in line lib/test_suite/two_fer.ex:1  Bad function true in line lib/test_suite/take_a_number.ex:1  Bad function true in line lib/test_suite/pacman_rules.ex:1
Last messages in the log cache:
  Reading files and computing callgraph... 
Halting VM with exit status 1
```

It looks very cryptic to me. Maybe there is a real problem in the code that needs fixing?

As far as I could find about the `@dialyzer` module attribute (http://erlang.org/doc/man/dialyzer.html#requesting-or-suppressing-warnings-in-source-files), it supports values listed here as `WarnOpts`: http://erlang.org/doc/man/dialyzer.html#gui-1

None of those seems to match "bad function" 🤔 so I think it wouldn't be possible to ignore this error via the `@dialyzer` module attribute.

Another option could be a `dialyzer_ignore.exs` file like we added to the track: https://github.com/exercism/elixir/blob/main/.dialyzer_ignore.exs (+ config https://github.com/exercism/elixir/blob/main/mix.exs#L12-L17). Assuming ElixirLS would use this file. I don't really use ElixirLS.

neenjaw:
Ya, it was a very obscure annotation when I found it. I should have written down the source. But it turned off the error and that is what mattered at the time. 😄

Ive never run dialyxir against the project directly because of it run implicitly through ElixirLS, so that error is a new one to me.

angelikatyborska:
Found this comment: https://github.com/elixir-gettext/gettext/issues/206#issuecomment-519805288 but it doesn't explicitly mention that `generated: true`  goes into `@dialyzer` 🤔

angelikatyborska:
After a few hours of working on this, I now have a theory...

> But it turned off the error and that is what mattered at the time. 😄

I think what happened, is that adding `@dialyzer generated: true` to the modules didn't resolve any dialyzer warnings, but it actually stopped dialyzer from being able to analyze the code in the first place. I guess that resulted in ElixirLS not reporting any dialyzer warnings.

Experiment that is supposed to prove this:

First, I added dialyzer to this project explicitly, also to the CI: [`9382ef4` (#58)](https://github.com/exercism/elixir-analyzer/pull/58/commits/9382ef46fa9c997d3a7ec9554daa46e73192f1aa)

The analysis on CI failed completely https://github.com/exercism/elixir-analyzer/runs/2421273594 with this weird error:
```
:dialyzer.run error: Analysis failed with error:
Could not scan the following file(s):
  Bad function true in line lib/test_suite/two_fer.ex:1  Bad function true in line lib/test_suite/take_a_number.ex:1  Bad function true in line lib/test_suite/pacman_rules.ex:1
```

Second, I remove all instances of `@dialyzer generated: true` (https://github.com/exercism/elixir-analyzer/pull/58/commits/8e3bad336f2cccf51f1e41ea1dd151eabc4fbffd). The CI now finishes the dialyzer analysis and reports warnings https://github.com/exercism/elixir-analyzer/runs/2421286712

Third, I fix one of the warnings because it's easy and ignore the second one because I have no idea how to fix it without rewriting everything (https://github.com/exercism/elixir-analyzer/pull/58/commits/bb66538c90c87a5b2f084962453cb2cd732212eb). The CI now passes :)

.dialyzer_ignore.exs
```diff
@@ -0,0 +1,18 @@
+[
+    ~r/The test \d* =:= nil can never evaluate to 'true'/
```
angelikatyborska, line 2: This is the offending line (when `depth` is used in the `feature` macro):
https://github.com/exercism/elixir-analyzer/blob/cc85fea1abaa1a113da7d787faefff1fc27e8217/lib/elixir_analyzer/exercise_test/feature/compiler.ex#L107

It fails because after unquoting in produces code `1 in [nil, depth]`. No idea how to fix this.

angelikatyborska: The goal I wanted to achieve in this PR is to remove the requirement to add this module attribute to every module that uses `ElixirAnalyzer.ExerciseTest` so that the process of writing an analyzer extension is a bit simpler.

If it was my project only, I would only remove this module attribute from everywhere and not do anything else. But since you're saying ElixirLS is a problem for you, and it will run dialyzer for every project, I think it would be nice to have a CI check that the dialyzer passes, even if the only advantage of that is that you (and other VSC users?) won't get bothered by warnings :).

So in my opinion, the PR in its current state satisfies that goal and I definitely don't need any rewrites of all the macros just to satisfy dialyzer 😅

neenjaw: Yea, without that attribute in each actual file, it fills my ide with "problems":

Screenshot from 2021-04-24 19-46-09

So if we could leave that in the test files I would appreciate it.  All other changes are fine.

angelikatyborska: Thank you for the screenshot, it made me realize where to look in VSC to see if my fix works. I can totally understand why you don't want to have those "problems" reported and I think I _FINALLY_ have a solution that works :).

However, I couldn't keep `@dialyzer generated: true` and also keep my other changes because this module attribute crashes the task `mix dialyzer`, I am 99% of that. With this attribute, the task `mix dialyzer` would always fail on CI.

*TL;DR:*
I realized that we cannot use dialyxir's ignore file to ignore the warnings because ElixirLS doesn't use dialyxir, it uses dialyzer directly.

I was able to ignore the specific warnings with correct `@dialyzer` annotations so that now ElixirLS doesn't report those "problems" in VSC, and also `mix dialyzer` passes.

I think we both have now what we wanted 🤝

neenjaw (approved):
Changes look good. Finally got some quiet time to actually settle in to read these PRs.  As always, excellent work!!

PR #59, opened 2021-04-23
Ensure all available test suites are assigned to exercises

angelikatyborska:
I would like to propose this test case that will check that all modules whose name start with `ElixirAnalyzer.TestSuite.` are referenced in the config that assigns exercises to analyzer modules. This would prevent us from merging PRs that add a new exercise but forget to add it to a config.

For this test to make sense, we would have to assume that we will not add unfinished test suites that shouldn't be run yet. That has been so far the case.


neenjaw (approved):
Seems reasonable, nice work!

PR #64, opened 2021-05-23
Add analyzer for german-sysadmin

angelikatyborska:
It checks that strings aren't used and that `case` is used.

Corresponding website copy PR: https://github.com/exercism/website-copy/pull/1992

### DIFF
--- a/lib/elixir_analyzer/test_suite/german_sysadmin.ex
+++ b/lib/elixir_analyzer/test_suite/german_sysadmin.ex
@@ -8,25 +8,25 @@ defmodule ElixirAnalyzer.TestSuite.GermanSysadmin do
   assert_no_call "doesn't convert anything to a string" do
     type :essential
     called_fn name: :to_string
-    comment ElixirAnalyzer.Constants.german_sysadmin_no_strings()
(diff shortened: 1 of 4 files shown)


lib/elixir_analyzer/test_suite/german_sysadmin.ex
```diff
@@ -0,0 +1,37 @@
...
+  assert_no_call "doesn't create binaries from character codes" do
+    type :essential
+    called_fn name: :<<>>
+    comment ElixirAnalyzer.Constants.german_sysadmin_no_strings()
+  end
+
+  assert_call "using case is required" do
+    type :essential
+    called_fn name: :case
+    comment ElixirAnalyzer.Constants.german_sysadmin_use_case()
```
neenjaw, line 35: Nice! This is a good example of how using `assert_call` works better than `feature` in some cases

neenjaw (approved):
Looks good!

lib/elixir_analyzer/test_suite/german_sysadmin.ex
angelikatyborska: Yup! I was really happy to realize it can be used like this.

PR #66, opened 2021-05-24
Raise when unknown expression used in feature/assert_call

### LINKED ISSUE #63: Warn / Error when directives in `feature` or `assert_call` tests are not supported
Both the `feature` and `assert_call` tests use a DSL to specify their success/error conditions.  When a directive is used that isn't supported, it fails silently (see recent PR comment [here](https://github.com/exercism/elixir-analyzer/pull/61#discussion_r637420056)). 

AC: I want to know when I am using a directive that is unsupported so that I can get better feedback when writing tests.

angelikatyborska:
Closes #63

### DIFF
--- a/lib/elixir_analyzer/exercise_test/assert_call.ex
+++ b/lib/elixir_analyzer/exercise_test/assert_call.ex
@@ -34,6 +34,8 @@ defmodule ElixirAnalyzer.ExerciseTest.AssertCall do
   end
 
   defp parse(description, block, should_call) do
+    :ok = validate_call_block(block, should_call)
+
     test_data =
       block
       |> walk_assert_call_block()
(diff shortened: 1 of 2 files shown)


lib/elixir_analyzer/exercise_test/assert_call.ex
```diff
@@ -34,6 +34,8 @@ defmodule ElixirAnalyzer.ExerciseTest.AssertCall do
   end
 
   defp parse(description, block, should_call) do
+    :ok = validate_call_block(block, should_call)
```
neenjaw, line 37: can we move this call to before lines `29` and `33`? Then we can pass the macro_name as an argument rather than using `should_call` and then coupling the macro_names in a deeper function?

neenjaw (changes requested):
Looks good, I think it should be up one more level though so that a deeper level doesn't need to know what the macro names are

lib/elixir_analyzer/exercise_test/assert_call.ex
angelikatyborska: Good idea: [`95b170a` (#66)](https://github.com/exercism/elixir-analyzer/pull/66/commits/95b170a2b5398fed085b1052875b4cffcc53094b)

neenjaw (approved):

PR #99, opened 2021-07-08
Mini refactoring

angelikatyborska:
I wanted to explore how to extend the functionality of the analyzer and found some opportunities for refactoring (IMO).

- Introduce a `Comment` struct that serves as documentation 
- Change `suppress_if` to a two tuple, mostly because you can't document a type like `[String.t(), :fail | :pass]`
- Rename some files and tests to remove any references to the old "status" and "referring to mentors"


angelikatyborska:
Something unexpected must have happened, a test is failing even though I shouldn't have changed anything that affects it 🙈

neenjaw:
`Error: 07-09T03:57:58.176Z [] [error] Loading exercise test suite '' failed (nofile).` I've never seen this error before, have you?

angelikatyborska:
Yes, that error is from the test case for `./test_data/two_fer/error_solution/` and it's expected. I added a `capture_log` call to avoid seeing this in the test output.

angelikatyborska:
I reverted the commit that caused the weird test failure and did all the same changes again, split in more commits. All tests pass :) I must have made a mistake when renaming directories.

neenjaw (approved):
Nice refactor adding the struct constraint to the map

PR #100, opened 2021-07-09
Change some moduledocs that have wrong descriptions

f-francine:
There are some modules that have a moduledoc  description a little wrong. For example, the following moduledoc is describing the Lasagna module.
```
@moduledoc """
  This is an exercise analyzer extension module for the concept exercise Take-A-Number
  """
```


neenjaw (approved):
Thanks!

Issue #97, opened 2021-07-08
Universal checks for every single solution?


angelikatyborska:
As I mentor, I keep seeing some reoccurring problems with solutions that are exercise-independent and could, at least partially, be detected automatically.

A code snippet from a real recent solution:
```elixir
@spec reverse(list) :: list
defp reverseHelper([], reversed) do
  reversed
end
defp reverseHelper([x|xs], reversed) do
  reverseHelper(xs, [x|reversed])
end

def reverse(l) do
  reverseHelper(l,[])
end
```

The analyzer could detect and give a suggestion to the user that:
- `@spec` for a function should be placed directly above that function definition, there shouldn't be a different function definition in between
- function (and variable) names should use snake_case

More opinionated, but we could also check if each public function has a `@spec` and `@doc` and nudge students to add them (for practice exercises only).

Without parsing the AST, we could check if the code is intended with spaces or with tabs.

I think in particular the camelCase and indenting are worth doing in the analyzer, because they're hard no-nos in community Elixir code, but they're annoying details that are pretty uncomfortable to point out to a new comer. It would be nicer if a machine did that 🙂.

neenjaw:
I like this idea in moderation.  The casing stuff I think is important enough to warrant extensions.  The indenting, I would rather point people to `mix format` rather than wade into a discussion of tabs vs spaces.

neenjaw:
We could create a mix format microservice (picoservice?) which accepts a chunk of code and "formats it" and returns it to the editor :P

angelikatyborska:
> The indenting, I would rather point people to mix format

In which way would you point people? From the context I'm guessing NOT by having an analyzer comment? Would you expect mentors to tell students that?

> We could create a mix format microservice (picoservice?) which accepts a chunk of code and "formats it" and returns it to the editor :P

That would be great because students in the web editor otherwise won't have access to the auto formatting :D but I have a hunch it wouldn't be part of v3.0...

Now that the base work for having common checks is in there, I'm going to create separate tickets for separate checks so that other people can work on them too.

angelikatyborska:
Added three so far that I was personally sure about. Feel free to add more or dispute specific ones 🙂 

https://github.com/exercism/elixir-analyzer/issues/103
https://github.com/exercism/elixir-analyzer/issues/104
https://github.com/exercism/elixir-analyzer/issues/105

jiegillet:
How about adding checks for `IO.inspect` and `IO.puts`?

[...]

PR #102, opened 2021-07-09
Run common checks for every single solution; Add common check for snake_case

angelikatyborska:
This is my attempt at adding common checks that will be done on every single solution, as suggested in https://github.com/exercism/elixir-analyzer/issues/97

For starters, I'm adding two checks that were pretty easy to implement:
- module attribute names should be snake_case
- function names should be snake_case

I can also see that it should be relatively easy to implement a check for PascalCase module names. It might be a bit more difficult to check for snake_case variable names, but still worth it.

I would also want to later add a check for not using tabs for indentation, but only as a `informative` comment not to be too annoying.

Some checks can be "stolen" from Credo :)
- module names: https://github.com/rrrene/credo/blob/master/lib/credo/check/readability/module_names.ex
- snake case variable names: https://github.com/rrrene/credo/blob/master/lib/credo/check/readability/variable_names.ex
- predicate function names (for Bob!): https://github.com/rrrene/credo/blob/master/lib/credo/check/readability/predicate_function_names.ex
- `@impl ModuleName` instead of `@impl true`: https://github.com/rrrene/credo/blob/master/lib/credo/check/readability/impl_true.ex

Website copy PR: https://github.com/exercism/website-copy/pull/2019


angelikatyborska:
@jiegillet your review would also be much appreciated!

lib/elixir_analyzer/test_suite/default.ex
```diff
@@ -0,0 +1,10 @@
+defmodule ElixirAnalyzer.TestSuite.Default do
+  @moduledoc """
+  This is the default exercise analyzer extension module.
+
+  It will be run for any exercise submission that doesn't have its own extension module.
+  It's empty, which means it will only run the common checks.
```
neenjaw, line 6:
```suggestion
  It's intentionally empty, which means it will only run the common checks.
```

neenjaw (approved):

lib/elixir_analyzer/exercise_test/common_checks/function_names.ex
```diff
@@ -0,0 +1,67 @@
...
+
+    if names == [] do
+      []
+    else
+      wrong_name =
+        names
+        |> Enum.reverse()
+        |> hd()
+        |> to_string()
+
```
neenjaw, line 28: Could be simplified a little

```suggestion
    wrong_name = List.last(names)
    
    if wrong_name do
      wrong_name = to_string(wrong_name)

```

lib/elixir_analyzer/exercise_test/common_checks/function_names.ex
```diff
@@ -0,0 +1,67 @@
...
+  end
+
+  defp snake_case?(name) do
+    # the code had to compile and pass all the tests to get to the analyzer
+    # so we can assume the name is otherwise valid
+    to_snake_case(name) == to_string(name)
+  end
+
+  defp to_snake_case(name) do
+    # Macro.underscore is good enough because a module attribute name must be a valid Elixir identifier anyway
```
neenjaw, line 64: 100% agree, despite the warning on the function not to use, I see it used everywhere in non-core code :P

lib/elixir_analyzer/exercise_test/common_checks/module_attribute_names.ex
```diff
@@ -0,0 +1,65 @@
...
+    {_, names} = Macro.prewalk(ast, [], &traverse/2)
+
+    if names == [] do
+      []
+    else
+      wrong_name =
+        names
+        |> Enum.reverse()
+        |> hd()
+        |> to_string()
```
neenjaw, line 25: Same simplification could be done here

lib/elixir_analyzer/exercise_test/common_checks/function_names.ex
```diff
@@ -0,0 +1,67 @@
...
+         %Comment{
+           type: :actionable,
+           comment: Constants.solution_function_name_snake_case(),
+           params: %{
+             expected: correct_name,
+             actual: wrong_name
+           }
+         }}
+      ]
+    end
```
neenjaw, line 42: Could be simplified a little

```suggestion
    wrong_name = List.last(names)
    
    if wrong_name do
      wrong_name = to_string(wrong_name)
      correct_name = to_snake_case(wrong_name)

      [
        {:fail,
         %Comment{
           type: :actionable,
           comment: Constants.solution_function_name_snake_case(),
           params: %{
             expected: correct_name,
             actual: wrong_name
           }
         }}
      ]
    else
      []
    end
```

angelikatyborska: `List.last`, why didn't I think of that 😂

angelikatyborska:
I'll merge to see it in action and add more checks in new PR(s)

PR #109, opened 2021-07-11
Unless with else

f-francine:


lib/elixir_analyzer/exercise_test/common_checks/unless_with_else.ex
```diff
@@ -0,0 +1,67 @@
...
+    end
+  end
+
+  defp else_block_for(args) when is_list(args) do
+    Enum.find_value(args, &find_keyword(&1, :else))
+  end
+
+  defp if_else_implementation({:unless, meta, args} = ast) do
+    [_, [do: do_block, else: else_block] = block] = args
+  end
```
f-francine, line 56: Here I intend to transform the unless/else implementation into a if/else one, so, my plan is to get the information in the else block and transform it into a if block, and transform the do_block into a else one, but I am not sure how can I do this. I would appreciate a little help here :)

neenjaw: OOC, Why do you need to transform it? What will you need to do with the transformed ast? I think the better first step for this would be to just identify it, and report it and the line it is on to the user as something to fix.  To fix and correct the `unless..else`into an `if..else` naively might not be straight forward and might lead to recommending bad practices.

For ex) What if a student wrote:
```elixir
unless not x do
  :a
else
  :b
end
```

Unless handled carefully, this might be corrected to:

```elixir
if not not x do
  :b
else
  :a
end
```

---

As a second step I think it would be really cool to transform it and then be able to return it to the user, but I think we need to identify common situations to avoid the above.  To answer the original question though, to reassemble the ast you just need to return it in the right format:

```elixir
  defp if_else_implementation({:unless, meta, args}) do
    [condition, [do: do_block, else: else_block]] = args
    inverted_condition = {:not, [context: Elixir, import: Kernel], [condition]}

    {:if, meta, [inverted_condition, [do: else_block, else: do_block]]}
  end
```

angelikatyborska: Transforming `unless/else` to `if/else` should be straight-forward because we don't need to touch the condition:
```
iex()> unless true, do: "false", else: "true"
"true"

iex()> if true, do: "true", else: "false"
"true"
```

However, I am against doing this (now or later) for two reasons:
- modified AST changed back to a string will have formatting that might drastically differ from what the student wrote in their code
- we have no idea how long the code blocks will be, vertically or horizontally, trying to use them in the analyzer comment might produce a comment that doesn't fit on the website and isn't very useful for explaining the issue with `unless/else`.

Let's just use a generic code example in the analyzer comment that will not have anything to do with the student's solution, but will show very well the problem 🙂

neenjaw:
How's it going @f-francine? Anything can do to help? No rush, just want to make sure you're not stuck

lib/elixir_analyzer/exercise_test/common_checks/unless_with_else.ex
f-francine: @angelikatyborska @neenjaw yeah, I didn't realize that tranforming the code to an if-else implementation would be problematic, but the points you guys raised really make sense.

f-francine:
> How's it going @f-francine? Anything can do to help? No rush, just want to make sure you're not stuck

Hey, @neenjaw! Sorry for the delay to awnser.
So, I do got stuck. I'm trying to following the credo's implementation, using
 ```elixir
traverse({:@, _, [{:unless, _, _}]} = ast, issues)
```
and 
```elixir
traverse({:unless, _, _} = ast, issues)
```
but I can't figure it out a way to make it works. I always get a match problem.

lib/elixir_analyzer/exercise_test/common_checks/unless_with_else.ex
```diff
@@ -0,0 +1,72 @@
...
+  defp traverse({:@, _, [{:unless, _, _}]} = ast, issues) do
+    if else_block?(ast) do
+      else_block = else_block_for(ast)
+      {ast, [else_block | issues]}
+    else
+      {ast, []}
+    end
+  end
+
+  defp traverse({:unless, _, _} = ast, issues) do
```
jiegillet, line 36: You are getting errors because you don't provide a default option for nodes that don't match either  `{:unless, _, _} ` or `{:@, _, [{:unless, _, _}]}` (which BTW I don't think you need since that would be matching code like `@unless x`). You need to add
```elixir
defp traverse(node, issues), do: node
```
Or something similar. Make sure you check code like
```elixir
quote do
  unless x, do: :a, else: :b
end
```
in `iex` to look at the structure of how `unless` works.

jiegillet:
Hi @f-francine,
If you are having trouble with handling the AST traversal, since #133, there is  another mechanism you could use to implement this common test. Now it could be as simple as something like (warning, I did not test this)
```elixir
defmodule ElixirAnalyzer.ExerciseTest.CommonChecks.UnlessWithElse do
  defmacro __using__(_opts) do
    quote do
      feature Constants.solution_unless_with_else() do
        comment Constants.solution_unless_with_else()
        form do
          unless _ignore do
            _ignore
          else
            _ignore
          end
        end
      end    
    end
  end
end
```
and add the the module here
https://github.com/exercism/elixir-analyzer/blob/ade6b518d97fb9fdb8756b9ca20ead6c12170649/lib/elixir_analyzer/exercise_test/common_checks.ex#L13-L18

jiegillet:
Hi @f-francine, we haven't heard from you in a while, so I would like to close the PR and get other people an opportunity to work on this. Of course, if you would like to continue working on it, you are more than welcome to do so.

Issue #104, opened 2021-07-10
Common check: variable names in snake_case


angelikatyborska:
Write a common check that is going to check variable names. They should always be written in snake_case.

See `lib/elixir_analyzer/exercise_test/common_checks.ex` and existing common checks for implementation details, e.g. https://github.com/exercism/elixir-analyzer/blob/e39b2cbe3f0565ff0b8f5ff6e3ccbf4826337479/lib/elixir_analyzer/exercise_test/common_checks/function_names.ex

This can be a bit complicated because variables can get created in different ways, e.g. on the left side of `=`, inside a function head, on the left size of `->` in a case statement, in an anonymous function... Use Credo's implementation of this check as inspiration: https://github.com/rrrene/credo/blob/master/lib/credo/check/readability/variable_names.ex

It doesn't have to be perfect. False-positives in this case are worse than false-negatives.

jiegillet:
I would like to try my hand at this one. 
I looked into it a bit, is there a reason why we can't simply match on
```
{name, _meta, module} when is_atom(module)
```
[This page says](https://elixir-lang.org/getting-started/meta/quote-and-unquote.html)

>The third element is either a list of arguments for the function call or an atom. When this element is an atom, it means the tuple represents a variable.

In any case, an extensive testing is probably good. Is it OK to ripoff `Credo`'s tests?

neenjaw:
Yes, I think you could do that match for the most part.  I don't know of a variable ast form off-hand that violates that match.

Credo is MIT licensed, so go for it :)

jiegillet:
OK, I'll see what passes the tests then.

I have a question, because I don't really know the history: is there a specific reason we are not using `Credo` itself? Seems like we might be re-inventing the wheel for some checks.

neenjaw:
Credo is more opinionated than we want to be overall.  Also this way we can return an intermediate form to build our final exercism-compatible form

PR #111, opened 2021-07-12
[#104] Common check: variable names in snake_case 

jiegillet:
Solves #104.

My code is a lot simpler than Credo, but passes the same test. Well, minus one that I commented:
```
... = parameter1  
```
For some reason, three periods is a valid variable name, but only three, not two, not four. Is that so you can use ellipsis in your code? Like `def f(x), do: ...`? Anyway`Macro.underscore` turns it into `./.` so it doesn't pass the test. I could make a special case for it, but it doesn't feel worth doing. Let me know what you think.


test/elixir_analyzer/exercise_test/common_checks/variable_names_test.exs
```diff
@@ -0,0 +1,419 @@
...
+          defmodule CredoSampleModule do
+            def some_function(parameter1, parameter2) do
+              some_value = parameter1 + parameter2
+              {some_value, _} = parameter1
+              [1, some_value] = parameter1
+              [some_value | tail] = parameter1
+              "e" <> some_value = parameter1
+              ^some_value = parameter1
+              %{some_value: some_value} = parameter1
+              #    ... = parameter1 # Ignore this pathological case
```
angelikatyborska, line 28: Since this is a single variable name that's so weird, let's add a case for it and make this pass :D it should be trivial, and I would want to avoid false-positives. Meanwhile I won't be able to sleep until I find out what is going on, why is this a valid variable name...

angelikatyborska (approved):
Great job finding the simplified pattern!

test/elixir_analyzer/exercise_test/common_checks/variable_names_test.exs
```diff
@@ -0,0 +1,419 @@
...
+              latency_μs = 5
```
angelikatyborska, line 29: no, wait, here we go, more cases that need to pass the test 😇 see https://github.com/rrrene/credo/blob/master/lib/credo/check/readability/variable_names.ex#L24
```suggestion
              latency_μs = 5
              user = %__MODULE__{}
              IO.inspect(__DIR__)
              IO.inspect(__ENV__)
            end

            defmacro foo() do
              IO.inspect(__CALLER__)
            end
```

lib/elixir_analyzer/exercise_test/common_checks/variable_names.ex
```diff
@@ -0,0 +1,63 @@
...
+             actual: wrong_name
+           }
+         }}
+      ]
+    else
+      []
+    end
+  end
+
+  defp traverse({name, _meta, module} = ast, names) when is_atom(module) do
```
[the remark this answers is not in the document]
jiegillet: Yes, I've been thinking about it more, and reached the same guess. But since the check is quite cheap, I wonder if there's really that much of a benefit. Not for us I would guess :)

test/elixir_analyzer/exercise_test/common_checks/variable_names_test.exs
jiegillet: Oh, nice catch, thank you!

Can I ask an unrelated question? I've been thinking about going into software/web development. If this PR was a professional one, how would it be received? Would you feel that I'm not doing my job right? Would you say it's OK from a junior developer only, or is this standard and expected that things like this will be missed? This is certainly showing me the necessity of peer review (and also that tests are hard).

angelikatyborska: I'll reply to this question on Slack, it feels a bit too personal to answer publicly 😅

Issue #78, opened 2021-06-30
Write a basic `square-root` analysis


jiegillet:
Write a check for the `square-root` exercise ([soon to be merged](https://github.com/exercism/elixir/pull/797) ) that makes sure `:math.sqrt`, `:math.pow`, `Float.pow` are not used.

https://github.com/exercism/problem-specifications/blob/main/exercises/square-root/description.md

Introduction to writing an analyzer extension: https://github.com/exercism/elixir-analyzer/blob/main/docs/writing-an-analyzer.md

An example of an analyzer extension that detect usages of disallowed modules: https://github.com/exercism/elixir-analyzer/pull/50/files#diff-39ba60165e48ed55ff5dc41ce966786ae8f1c83abf1cc3e2430e0b8b5692f105

jiegillet:
I'll probably start on this one after #104 is done. It is mine after all :)

jiegillet:
This can be closed BTW

PR #112, opened 2021-07-13
[#78] Write a basic square-root analysis

jiegillet:
Solves #78.

This is possibly still WIP, I need some advice.

Here, I want to specifically forbid `:math.pow`, `:math.sqrt` and `Float.pow`, which I could do, but to prevent students from importing/aliasing, I did not find a better way than to ban the whole `Float` and `:math` packages (BTW, `assert_call` would not recognize Erlang libraries so I fixed that).

Admittedly, student probably don't need other functions in `Float` or `:math` either, but in case they do, is there a more subtle way to do this? Maybe annotating the aliases and imports as the nodes are traversed? Is this worth doing? I could more or less follow the code in `assert_call` so I could possibly try...


.gitignore
```diff
@@ -36,3 +36,6 @@ tmp/
 
 /priv/plts/*.plt
 /priv/plts/*.plt.hash
+
+# Vim temporary files
+*.swp
```
angelikatyborska, line 41: I don't mind this addition at all, but since you're considering a career in dev, I'll give you this feedback: usually files that need to be ignored because of your specific tooling preferences don't go in the project's `.gitignore`, but your own global `.gitignore`. See: https://sebastiandedeyne.com/setting-up-a-global-gitignore-file/

My personal `.gitignore` looks like this (macOS and RubyMine stuff)
```
$ cat ~/.gitignore
.DS_Store
.AppleDouble
.LSOverride

# Icon must end with two \r
Icon

# Thumbnails
._*

# Files that might appear in the root of a volume
.DocumentRevisions-V100
.fseventsd
.Spotlight-V100
.TemporaryItems
.Trashes
.VolumeIcon.icns

# Directories potentially created on remote AFP share
.AppleDB
.AppleDesktop
Network Trash Folder
Temporary Items
.apdisk

.idea
*.iml
```

lib/elixir_analyzer/exercise_test/assert_call/compiler.ex
```diff
@@ -139,13 +139,29 @@ defmodule ElixirAnalyzer.ExerciseTest.AssertCall.Compiler do
...
   end
 
+  # For erlang libraries: :math.pow
+  def matching_function_call?(
+        {{:., _, [module_path, name]}, _, _args},
+        {module_path, name}
+      ) do
+    true
+  end
+
```
angelikatyborska, line 164: Could you add two new test cases to `test/elixir_analyzer/exercise_test/assert_call_test.exs` to test this addition?

angelikatyborska (commented):
>  (BTW, assert_call would not recognize Erlang libraries so I fixed that).

👍👍👍 

> Admittedly, student probably don't need other functions in Float or :math either

Even if they don't need them, I would rather not forbid the whole module 🤔 We don't need to be super strict with forbidding function usage. If somebody _wants to_ walk go around it by importing the whole module, whatever 🤷 we're not a school with grades and tuition :)

> Maybe annotating the aliases and imports as the nodes are traversed? Is this worth doing? 

I have a hunch that you would be interested in trying. If that's true, I'd say go ahead. It would be really cool if `assert_call` could actually correctly follow aliases and imports, but it's optional IMO.

.gitignore
jiegillet: OMG is so obvious. Of course I should have been doing this for years!  :hurtrealbad: :hurtrealbad: :hurtrealbad:

lib/elixir_analyzer/exercise_test/assert_call/compiler.ex
jiegillet: Excellent idea.

jiegillet:
Oh no, I forgot I had already started the PR. Don't look at the last commit, it doesn't even compile, it was a I'm-switching-computer-gotta-save-my-work kind of commit.

jiegillet:
I'm not totally done, but this broken commit was a weight on my soul, so I'm pushing some progress.

I managed to track imported and aliased functions 🥳 
More things I need to do
* I want to be able to detect modules imported and aliased inside of functions, should not be too hard since there is already code to keep track of functions 
* Add tests, many many tests ^^
* Fix Credo warnings (I gotta start using that more. Is there a systematic way to run it?)
* Refactor? 
* PR on website-copy

jiegillet:
My above checklist is done. 
I've added a bunch of tests, but I probably did not do that well, as I sometimes highjacked some existing test to test two things at once (especially with the `A.B.C.efg()` tests in `test/elixir_analyzer/exercise_test/assert_call_test.exs`) so I'm open to suggestions on how to structure all that better.

neenjaw:
> My above checklist is done.
> I've added a bunch of tests, but I probably did not do that well, as I sometimes highjacked some existing test to test two things at once (especially with the `A.B.C.efg()` tests in `test/elixir_analyzer/exercise_test/assert_call_test.exs`) so I'm open to suggestions on how to structure all that better.

I'll take a look, haven't really looked at your approach to tracking the imports yet.

> Also, would it be better to split this PR into two? One for improvements on `asset_call` and one for square root.

Maybe in the future, for this don't worry about it. I don't believe in the commit/pr record having to be perfect.

lib/elixir_analyzer/exercise_test/assert_call/compiler.ex
```diff
@@ -94,12 +104,27 @@ defmodule ElixirAnalyzer.ExerciseTest.AssertCall.Compiler do
...
+    do: get_paths(module_path, :import)
+
+  def add_directive({:alias, _, [module_path]}),
+    do: get_paths(module_path, :alias)
+
+  def add_directive({:alias, _, [module_path, [as: {:__aliases__, _, [alias]}]]}),
+    do: get_paths(module_path, :alias) |> Enum.map(fn {_, path} -> {[alias], path} end)
+
+  def add_directive(_), do: []
+
```
neenjaw, line 127: I like the idea of this function bu I have a couple suggestions:
1) this function is doing two things, which might be why the name of the function `add_directive` doesn't seem to quite fit. It is detecting and tracking imports and detecting and tracking aliases, I think, while similar, it is worthwhile to handle them separately
   - Suggestion: split these up even if it isn't as DRY if it makes more sense

     ```elixir
     # I think it would flow well if it could be used in the annotate function like this to replace lines 79-85:
     acc =
       node
       |> trackImports(node, acc)
       |> trackAliases(node, acc)
     ```

2) it seems to depend on the reducer in the `annotate/2` function on line 81-85, so that should probably be inlined into this function as a private helper function

lib/elixir_analyzer/exercise_test/assert_call/compiler.ex
```diff
@@ -195,4 +244,35 @@ defmodule ElixirAnalyzer.ExerciseTest.AssertCall.Compiler do
...
+    [{_, root_path}] = get_paths(root, :alias)
+
+    for branch <- branches,
+        {last, full_path} <- get_paths(branch, :alias) do
+      {last, root_path ++ full_path}
+    end
+  end
+
+  def get_paths(path, _type) when is_atom(path),
+    do: [{[path], [path]}]
```
neenjaw, line 277: This, like above, seems to service two use cases and I think they are being merged together because they are similar, but they are a little bit different and that makes them harder to understand since it is hard to know if they can jump between the two (I suspect they can't but it appears that way on inspection).  Suggestion: separate them into two function `get_import_paths` and `get_alias_paths`

neenjaw (changes requested):
Left some suggestions about your additions to track the import and aliases.  Like the approach, but I think we can make the code easier to read/understand.

Everything else seems good to go.

lib/elixir_analyzer/exercise_test/assert_call/compiler.ex
jiegillet:
```
trackImports
```
Hahaha, have you been coding in... mmh... JavaScript maybe?

neenjaw: My daily work is all JavaScript and PHP, so yep

lib/elixir_analyzer/exercise_test/assert_call/compiler.ex
```diff
@@ -74,6 +76,15 @@ defmodule ElixirAnalyzer.ExerciseTest.AssertCall.Compiler do
...
   @spec annotate(Macro.t(), map()) :: {Macro.t(), map()}
   def annotate(node, acc) do
+    imports = track_imports(node) ++ track_aliases(node)
+
+    acc =
+      Enum.reduce(imports, acc, fn {alias, full_path}, acc ->
+        if acc.in_function_def,
+          do: %{acc | in_function_modules: Map.put(acc.in_function_modules, alias, full_path)},
+          else: %{acc | modules_in_scope: Map.put(acc.modules_in_scope, alias, full_path)}
+      end)
```
neenjaw, line 86: I dont think annotate should be responsible for folding the results of `track_*` into the accumulator, I have included a further code suggestion with passing code below.

```suggestion
    acc =
      acc
      |> track_aliases(node)
      |> track_imports(node)
```

lib/elixir_analyzer/exercise_test/assert_call/compiler.ex
```diff
@@ -195,4 +230,47 @@ defmodule ElixirAnalyzer.ExerciseTest.AssertCall.Compiler do
...
+    [{_, root_path}] = get_alias_paths(root)
+
+    for branch <- branches,
+        {last, full_path} <- get_alias_paths(branch) do
+      {last, root_path ++ full_path}
+    end
+  end
+
+  defp get_alias_paths(path) when is_atom(path),
+    do: [{[path], [path]}]
```
neenjaw, line 275: Following the suggestion above to extract the`Enum.reduce` to a private function to be reused by `track_import/2` and `track_alias/2`.  I would prefer this style  because it means to understand `annotate` we don't also have to decipher the `Enum.reduce` expression now found in `track_modules/2`.

Also, while I may not be 100% accurate in my own coding, if one function clause has required the multi-line function block once, they should all be multiline blocks for consistency.

```suggestion
  # track_imports
  defp track_imports(acc, {:import, _, [module_path | _]}) do
    paths = get_import_paths(module_path)
    track_modules(acc, paths)
  end

  defp track_imports(acc, _) do
    acc
  end

  # get_import_paths
  defp get_import_paths({:__aliases__, _, path}) do
    [{path, path}]
  end

  defp get_import_paths({{:., _, [root, :{}]}, _, branches}) do
    [{root_path, _}] = get_import_paths(root)

    for branch <- branches,
        {path, _} <- get_import_paths(branch) do
      {root_path ++ path, root_path ++ path}
    end
  end

  defp get_import_paths(path) when is_atom(path) do
    [{[path], [path]}]
  end

  # track_aliases
  defp track_aliases(acc, {:alias, _, [module_path]}) do
    paths = get_alias_paths(module_path)
    track_modules(acc, paths)
  end

  defp track_aliases(acc, {:alias, _, [module_path, [as: {:__aliases__, _, [alias]}]]}) do
    paths = get_alias_paths(module_path) |> Enum.map(fn {_, path} -> {[alias], path} end)
    track_modules(acc, paths)
  end

  defp track_aliases(acc, _) do
    acc
  end

  # get_alias_paths
  defp get_alias_paths({:__aliases__, _, path}) do
    [{[List.last(path)], path}]
  end

  defp get_alias_paths({{:., _, [root, :{}]}, _, branches}) do
    [{_, root_path}] = get_alias_paths(root)

    for branch <- branches,
        {last, full_path} <- get_alias_paths(branch) do
      {last, root_path ++ full_path}
    end
  end

  defp get_alias_paths(path) when is_atom(path) do
    [{[path], [path]}]
  end

  # track modules
  defp track_modules(acc, module_paths) do
    Enum.reduce(module_paths, acc, fn {alias, full_path}, acc ->
      if acc.in_function_def,
        do: %{acc | in_function_modules: Map.put(acc.in_function_modules, alias, full_path)},
        else: %{acc | modules_in_scope: Map.put(acc.modules_in_scope, alias, full_path)}
    end)
  end
```

neenjaw (changes requested):
Good iteration, I think we can go further to ease the responsibility of the `annotate` function.  See code suggestions.  I have tested them, they pass.

jiegillet:
Ah yes, I see now that's what you were suggesting in your first review. Sorry I missed it and thank you for adding suggestions.

angelikatyborska (commented):
I'm not exactly happy with how the `assert_(no_)call_test` tests are shaping up... it's not your fault, we set them up to fail by putting everything together 😅. It's very hard to tell from the new additions to the test files if the new features are properly tested. I would like to propose:
- create separate files for the two changes (detecting calls of erlang functions and keeping track of imports and aliases), e.g. `test/elixir_analyzer/exercise_test/assert_call/erlang_modules_test.exs` and `test/support/analyzer_verification/assert_call/erlang_modules.ex`
- You can mix up testing `assert_call` and `assert_no_call` in those new test files if that will make it easier.
- Make sure to also implement and test `:only` and `:except` options of import. E.g. if I `import String, except: [downcase: 1]` and then define and use my local `downcase/1` function, that shouldn't be mistaken for a `String.downcase` call. I don't see any tests for that at the moment 🤔 
- Please use more meaningful names in the tests than `A.B.C`, for example real modules and function from the standard library. It will be easier to follow the code if it's less abstract.

neenjaw:
> Ah yes, I see now that's what you were suggesting in your first review. Sorry I missed it and thank you for adding suggestions.

You're welcome, I saw that you tried to implement what I had suggested initially and I wasn't sure how to word it to further refine the approach.  Hope it didn't step on your toes too much to provide the suggestion. :)

neenjaw (approved):

neenjaw:
Once @angelikatyborska is happy with the tests, I think this is ready

lib/elixir_analyzer/exercise_test/assert_call.ex
```diff
@@ -133,14 +133,16 @@ defmodule ElixirAnalyzer.ExerciseTest.AssertCall do
...
+
         {:__aliases__, _, module} ->
           module
 
         x ->
           raise ArgumentError,
-                "calling function signature requires :module to be nil or a module atom, got: #{
-                  inspect(x)
-                }"
+                "calling function signature requires :module to be nil or a module atom, got: #{inspect(x)}"
```
angelikatyborska, line 145: (This is general Elixir dev feedback, not related to the feature)

This formatting change tells me that you're using Elixir 1.12, which in turn tells me you're not using asdf (https://asdf-vm.com/#/) to manage Elixir versions. I know this because if you were using asdf, it would automatically read `.tool-versions` and switch to the appropriate version, and there was a change in how strings get formatted between 1.11 and 1.12 😁 

We forgot to update this repository to Elixir 1.12 when we updated the other ones 🙈. You can see in the file `.tool-versions` (and also by the Docker container used in the Github Action).

Anyway, you don't need to change anything. I'm going to open PRs that updates Elixir here to 1.12 and that does the formatting check on CI to catch this automatically.

jiegillet: Elementary, my dear Watson. Nice deductions :)

I don't know asdf, I'm looking at the webpage. Do you recommend it?

jiegillet:
> You're welcome, I saw that you tried to implement what I had suggested initially and I wasn't sure how to word it to further refine the approach. Hope it didn't step on your toes too much to provide the suggestion. :)

Not at all, I appreciate it, communicating is hard, especially for this kind of abstract things.
I'll try to add this principle to my list of good practice. I don't very often share codebases with other people, so this is valuable.

lib/elixir_analyzer/exercise_test/assert_call.ex
neenjaw: If you are already using homebrew or something to manage versions, no.  If you aren't using anything and are installing them directly on to your machine, yes.

angelikatyborska: asdf is excellent for synchronizing dependency versions in teams, especially when there's more than one to manage (Erlang, Elixir, NodeJS). I definitely recommend it if you're working on different projects that require different versions of the same language. If not, then it's a nice to have but won't be gamechanging.

I have bad experiences with `homebrew`. Basically any time I install any package, it updates random unrelated packages which sometimes leads to things breaking unexpectedly. For example installing PHP broke my Postgres installation once 🤷

jiegillet:
I realized that I did not consider the `:only` and `:except` conditions before, so I added that.

I decided to treat `import Module, only: :functions` and `import Module, only: :macros` as ``import Module` because we don't have a general way to check if something is a macro or  function from where it is called. Hopefully that's not a big deal?

Writing tests is mind boggling, I never quite know when to stop or which combinations of things to test... I hope I did OK.

lib/elixir_analyzer/exercise_test/assert_call.ex
jiegillet: Yeah, I think homebrew broke my Postgres too once. I see the need for asdf and I will probably try it out at some point.

angelikatyborska:
```
I decided to treat import Module, only: :functions and import Module, only: :macros as import Module because we don't have a general way to check if something is a macro or function from where it is called. Hopefully that's not a big deal?
```

That makes sense to me, those two options are very rare in my experience.

test/support/analyzer_verification/assert_call/module_tracking.ex
```diff
@@ -0,0 +1,34 @@
...
+    called_fn module: Elixir.Mix.Utils, name: :read_path
+    comment "found a call to Elixir.Mix.Utils.read_path/1"
+  end
+
+  #  assert_no_call "didn't call to Elixir.Mix.Utils.read_path/1 in function/0" do
+  #    type :informational
+  #    called_fn module: Elixir.Mix.Utils, name: :read_path
+  #    calling_fn module: AssertCallVerification, name: :function
+  #    comment "found a call to Elixir.Mix.Utils.read_path/1 in function/0"
+  #  end
```
angelikatyborska, line 33: What's up with the commented code? Is it some work in progress or trash?

angelikatyborska:
I took a short look so far (I'm low on time Monday through Wednesday). I expected to be able to slim down some of the existing analyzer extensions for the exercises that forbid the usage of `Enum` for example. I made this commit and I expected all the tests to be passing, but they don't pass: https://github.com/jiegillet/elixir-analyzer/compare/jie-square-root...exercism:jie-square-root-test

Am I doing something wrong? Or is there maybe a bug in the tracking of imports and aliases?

test/support/analyzer_verification/assert_call/module_tracking.ex
jiegillet: Or sorry about that. I need to remove it.

jiegillet:
> Am I doing something wrong? Or is there maybe a bug in the tracking of imports and aliases?

I took a look, it's no so much a bug as a limitation of the method.
On test that fails is
```elixir
defmodule BirdCount do
  import Enum
  def total(list), do: sum(list)
end
```
It fails because `assert_no_call` doesn't know that `sum` is an `Enum` function. It can only check for `Enum._` functions or function names that were specified. If you specified here that `sum` is not to be used, it would work, but it won't work for all the `Enum` functions. 

I suppose one way to achieve that would be to actually import all the functions available in Enum, using `Enum.__info__(:functions)` (which I just googled). Actually then I can also do `Enum.__info__(:macros)`. There is an Erlang equivalent too: `:math.module_info(:exports)`. 

Shall I do that? This "good first issue" keeps growing haha :D

neenjaw:
> I suppose one way to achieve that would be to actually import all the functions available in Enum, using `Enum.__info__(:functions)` (which I just googled). Actually then I can also do `Enum.__info__(:macros)`. There is an Erlang equivalent too: `:math.module_info(:exports)`.
> 
> Shall I do that? This "good first issue" keeps growing haha :D

I think it is certainly possible, BUT the rabbit hole is indeed endless for some of these things and I think probably for a first step we should just draw a line and say at this version, it does _____ and has _____ limitations.  Then plan for next and further steps and evaluate what we actually need. I'd rather implement a function that helps us with 90% of our analyzer cases than one that helps with 1% of them. The analyzer should be used for heavy/monotonous cases where simple actionable comments can be provided to students and free up mentors to provide feedback and challenge for better software design.

Also, what matters to me is that we enjoy the time we put into this project, and, if it is becoming a slog or a "never-ending" ticket, we should tidy it up, merge it, and then do something else.

On a side note, one thing that I maybe regret is not giving more feedback on your example solutions as you were creating practice exercises.  Not that your code is bad, you are certainly talented, but I feel like I dropped the ball as someone here to collaborate with.

angelikatyborska:
I totally agree with Tim on:
> Also, what matters to me is that we enjoy the time we put into this project, and, if it is becoming a slog or a "never-ending" ticket, we should tidy it up, merge it, and then do something else.

and because of that, my answer to the question

> Shall I do that? This "good first issue" keeps growing haha :D

is: only if you feel like it 🙂 we can merge this PR as is (after I review the tests again) and we can let it be like this, or you can also do it in a separate PR later if this one is getting too big.

If you do:
> Enum.__info__(:functions)

keep in mind that this will only work for the standard library (which is probably 99.9% cases of imports that we will get). We cannot do that for other modules defined by the student because the code is not compiled.

jiegillet:
Truth is... I already have a version that seems to work 😅 
However it's a bit messy and I'm refactoring it. And crucially I need to add tests, because I've been bit by rushing too much. 
If you review the tests before I'm happy (take your time BTW), go ahead and merge the PR as is, I'll do another PR.

This has not become a slog, I quite enjoy it, so no worries about that :)

> On a side note, one thing that I maybe regret is not giving more feedback on your example solutions as you were creating practice exercises. Not that your code is bad, you are certainly talented, but I feel like I dropped the ball as someone here to collaborate with.

I'm touched by that thought, thank you. You don't need to feel bad at all, and I'll look forward to future collaborations.

jiegillet:
> keep in mind that this will only work for the standard library (which is probably 99.9% cases of imports that we will get). We cannot do that for other modules defined by the student because the code is not compiled.

99.9% should be good enough I think. 
However, that raises a question, what happens when a check crashes? Does the student still see the other checks?

jiegillet:
Well, I ended up committing the extra feature here, including @angelikatyborska's extra tests and some more of my own. 

The main change is in `track_imports` and `matching_function_call?` got a bit simpler in the process, nice bonus.

test/elixir_analyzer/exercise_test/assert_call/erlang_modules_test.exs
```diff
@@ -0,0 +1,154 @@
...
+      end,
+      defmodule AssertCallVerification do
+        alias :rand, as: Rand
+
+        def function() do
+          r = Rand.normal()
+        end
+      end,
+      defmodule AssertCallVerification do
+        alias :rand
```
angelikatyborska, line 26: lol 😂 why not

test/elixir_analyzer/exercise_test/assert_call/erlang_modules_test.exs
```diff
@@ -0,0 +1,154 @@
...
+
+        def function() do
+          r = :rand.normal()
+        end
+      end,
+      defmodule AssertCallVerification do
+        def function() do
+          r = :rand.normal()
+        end
+      end
```
angelikatyborska, line 36: Isn't this snippet exactly the same as the first one in this list?

test/elixir_analyzer/exercise_test/assert_call/erlang_modules_test.exs
```diff
@@ -0,0 +1,154 @@
...
+    ]
+  end
+
+  test_exercise_analysis "Searching for use of module with imported module will suceed",
```
angelikatyborska, line 40:
```suggestion
  test_exercise_analysis "Searching for use of module with imported module will succeed",
```

test/elixir_analyzer/exercise_test/assert_call/module_tracking_test.exs
```diff
@@ -0,0 +1,335 @@
...
+
+        def function() do
+          {:ok, file} = E.Mix.Utils.read_path("mix.exs")
+        end
+      end,
+      defmodule AssertCallVerification do
+        alias Elixir.Mix.Utils, as: U
+
+        def function() do
+          {:ok, file} = Elixir.Mix.Utils.read_path("mix.exs")
```
angelikatyborska, line 138: Was this meant to use the alias?
```suggestion
        alias Elixir.Mix.Utils, as: U

        def function() do
          {:ok, file} = U.Mix.Utils.read_path("mix.exs")
```

angelikatyborska (approved):
Brilliant! And the tests are so beautiful now 😍. Amazing job!

@neenjaw if you also like the new changes, can you merge?

PS: I pushed a small commit with updated documentation.

neenjaw:
Yes, I can look it over tonight

test/elixir_analyzer/exercise_test/assert_call/erlang_modules_test.exs
jiegillet: Just throwing anything I could think of :)

test/elixir_analyzer/exercise_test/assert_call/erlang_modules_test.exs
jiegillet: Indeed, I'll get rid of it. Maybe it was something else before I made changes to `import`.

lib/elixir_analyzer/exercise_test/assert_call/compiler.ex
```diff
@@ -231,20 +227,90 @@ defmodule ElixirAnalyzer.ExerciseTest.AssertCall.Compiler do
 
   def extract_function_name(_), do: nil
 
+  @doc """
+  Construct a callable module name from a list of atoms or a single atom for Erlang modules
+  """
+  def reconstruct_module_path([Elixir | _] = atoms) do
+    atoms
+    |> Enum.map_join(".", &to_string/1)
+    |> String.to_atom()
```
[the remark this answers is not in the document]
jiegillet: Sweet! In hindsight, of course this exists, but thanks for showing me :)

test/elixir_analyzer/exercise_test/assert_call/module_tracking_test.exs
jiegillet: No, here I meant to import an alias and check if the full path still works.

PR #113, opened 2021-07-17
Fix variable name check reporting function names

angelikatyborska:
I discovered this issue accidentally by playing around. Apparently a function definition without parenthesis produces a different AST then exactly the same function definition with parenthesis:

```elixir
iex(1)> x = quote do
...(1)> def foo, do: nil
...(1)> end
{:def, [context: Elixir, import: Kernel],
 [{:foo, [context: Elixir], Elixir}, [do: nil]]}

iex(2)> y = quote do
...(2)> def foo(), do: nil
...(2)> end
{:def, [context: Elixir, import: Kernel],
 [{:foo, [context: Elixir], []}, [do: nil]]}
```

I don't know how to fix this yet... I opened a PR with failing tests as a reminder.

### DIFF
--- a/lib/elixir_analyzer/exercise_test/common_checks/variable_names.ex
+++ b/lib/elixir_analyzer/exercise_test/common_checks/variable_names.ex
@@ -13,7 +13,10 @@ defmodule ElixirAnalyzer.ExerciseTest.CommonChecks.VariableNames do
 
   @spec run(Macro.t()) :: [{:pass | :fail | :skip, %Comment{}}]
   def run(ast) do
-    {_, names} = Macro.prewalk(ast, [], &traverse/2)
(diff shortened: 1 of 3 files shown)


jiegillet:
Oops, that's a bug in my code :(
We could make the variable name checks be context sensitive, like Credo does. 
Another (hacky?) way could be to send a warning about missing parentheses to function definitions that suppress the variable name check.
Usually `mix format` would add those parentheses, but not online I guess.

angelikatyborska:
@jiegillet I went for a more direct approach of detecting and ignoring this one corner case 🙂 doing it the Credo way, with specifying which operators can define variables and on which side... that's just so much work 😅

jiegillet (approved):
Yep, seems to work. And yes, WAY less work than the Credo way :)

neenjaw (approved):

Issue #88, opened 2021-07-07
Write analysis for concept exercise `name-badge`


angelikatyborska:
Desired checks: https://github.com/exercism/elixir/blob/9774308bb6b898df6ecbde64c5c6e48c651870ae/exercises/concept/name-badge/.meta/design.md#L23-L25

jiegillet:
I will do this one since it's quick and easy :)

PR #116, opened 2021-07-24
[#88] Write analysis for concept exercise `name-badge`

jiegillet:
Solves #88.
I picked a simple one this time for a change.

### DIFF
--- a/lib/elixir_analyzer/test_suite/name_badge.ex
+++ b/lib/elixir_analyzer/test_suite/name_badge.ex
@@ -7,8 +7,13 @@ defmodule ElixirAnalyzer.TestSuite.NameBadge do
   alias ElixirAnalyzer.Constants
 
   feature "Use an if macro" do
+    find :one
     type :essential
     comment Constants.name_badge_use_if()
+
+    form do
+      if(_ignore, do: _ignore)
+    end
 
     form do
       if(_ignore, do: _ignore, else: _ignore)
--- a/lib/elixir_analyzer/test_suite/name_badge.ex
+++ b/lib/elixir_analyzer/test_suite/name_badge.ex
@@ -1,0 +1,17 @@
+defmodule ElixirAnalyzer.TestSuite.NameBadge do
+  @moduledoc """
+  This is an exercise analyzer extension module for the concept exercise NameBadge
+  """
+
+  use ElixirAnalyzer.ExerciseTest
+  alias ElixirAnalyzer.Constants
+
+  feature "Use an if macro" do
+    type :essential
+    comment Constants.name_badge_use_if()
+
+    form do
+      if(_ignore, do: _ignore, else: _ignore)
+    end
+  end
+end
--- a/lib/elixir_analyzer/test_suite/name_badge.ex
+++ b/lib/elixir_analyzer/test_suite/name_badge.ex
@@ -7,7 +7,7 @@ defmodule ElixirAnalyzer.TestSuite.NameBadge do
   alias ElixirAnalyzer.Constants
 
   feature "Use an if macro" do
-    find :one
+    find :any
     type :essential
     comment Constants.name_badge_use_if()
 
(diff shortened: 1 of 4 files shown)


lib/elixir_analyzer/test_suite/name_badge.ex
```diff
@@ -0,0 +1,17 @@
...
+
+  use ElixirAnalyzer.ExerciseTest
+  alias ElixirAnalyzer.Constants
+
+  feature "Use an if macro" do
+    type :essential
+    comment Constants.name_badge_use_if()
+
+    form do
+      if(_ignore, do: _ignore, else: _ignore)
```
angelikatyborska, line 14: I wonder if we should allow either `if` with `do` or `if` with `do` and `else`? It's probably pretty hard to solve it with just `if`s without `else`s, but I worry about user frustration because the comment (that I just merged...) doesn't say that you need an `else` for this check to pass 😁

jiegillet: I actually started with a version that allowed `if` without `else` but ended up getting rid of it at the last minute (I didn't have good tests for it). I'll put it back in then :)

jiegillet:
Side questions about the test for the exercises, it's certainly best of they are valid solutions but do they need to be? I have not actually tested all of my illegal-but-should-still-pass solutions. Or is the real purpose of these tests is to add real use case to `feature` and `assert_call` tests?

lib/elixir_analyzer/test_suite/name_badge.ex
```diff
@@ -0,0 +1,22 @@
...
+  @moduledoc """
+  This is an exercise analyzer extension module for the concept exercise NameBadge
+  """
+
+  use ElixirAnalyzer.ExerciseTest
+  alias ElixirAnalyzer.Constants
+
+  feature "Use an if macro" do
+    find :one
+    type :essential
```
neenjaw, line 11: This is another case where I think this is too strong of a suggestion.

```suggestion
    type :actionable
```

angelikatyborska: But this is a learning exercise. I really think it should remain `essential`.

angelikatyborska:
> it's certainly best of they are valid solutions but do they need to be? 

They don't need to be. Most of them should so that they serve as a justification for the check ("look, there is this valid solution that somebody might come up with, we should check this one in particular"), but they don't need to be because human creativity is both limited (on an individual level) and limitless (on a group level). That means you might want to add a check for some weird corner case that you personally wouldn't know how to add to a valid solution, but nonetheless it's worth checking.

lib/elixir_analyzer/test_suite/name_badge.ex
angelikatyborska: I'll see on Slack what others think about this.

jiegillet:
> human creativity is both limited (on an individual level)

Speak for yourself ;)

Thank you, that makes sense.

lib/elixir_analyzer/test_suite/name_badge.ex
```diff
@@ -0,0 +1,22 @@
+defmodule ElixirAnalyzer.TestSuite.NameBadge do
+  @moduledoc """
+  This is an exercise analyzer extension module for the concept exercise NameBadge
+  """
+
+  use ElixirAnalyzer.ExerciseTest
+  alias ElixirAnalyzer.Constants
+
+  feature "Use an if macro" do
+    find :one
```
angelikatyborska, line 10: AFAIK `:one` is not a valid `find` value
```suggestion
    find :any
```

jiegillet: It is a valid value according to the [docs](https://github.com/exercism/elixir-analyzer/blob/main/docs/step-03/step-03.md) (line 208, GitHub is weird for linking markdown files I guess) but is means "match only one case" so `:any` is the best choice. Thank you.

angelikatyborska: Oops, my bad :D I haven't seen it before.

angelikatyborska (approved):

PR #117, opened 2021-07-24
Add common checks for tabs in indentation

angelikatyborska:
Website copy: https://github.com/exercism/website-copy/pull/2031

A very simple check that will detect if at least one tab was used somewhere as either leading whitespace (indentation) or trailing whitespace.


jiegillet:
Very nice. We can also use this technique to check for use use of multiline strings.

lib/elixir_analyzer/exercise_test/common_checks/indentation.ex
```diff
@@ -0,0 +1,27 @@
...
+           comment: Constants.solution_indentation()
+         }}
+      ]
+    else
+      []
+    end
+  end
+
+  defp uses_tabs_for_indentation?(ast, string) do
+    String.contains?(string, "\t") && !String.contains?(Macro.to_string(ast), "\\t")
```
jiegillet, line 25: What if tabs are used for indentation AND is used in strings?

jiegillet: Maybe you could count the tabs in `string` and that should match the numbers of tabs in `Macro.to_string(ast)`.

neenjaw: I think this is a good point, what if a tab appears in a string literal which is intended to be there and is not being used for indentation, will this give a false-positive error?

Maybe a safer approach is:
```elixir
tab_in_leading_whitespace = ~r/^\s*\t/

string
|> String.split("\n")
|> Enum.any(&(String.match?(&1, tab_in_leading_whitespace)))
```

jiegillet: This will trigger a fail when a multiline string starts with tabs, I think.

neenjaw: Ok, then more strictly:

```elixir
tab_in_leading_whitespace = ~r/^\t/
```

jiegillet: This would miss `\s\t\s` which is arguably a way worse indentation style :D
I think counting the number of "intended" tabs and making sure that that is the only tabs in the source would work well.

neenjaw: It would also miss `\s\s\t\s` -- my point was to remove the false-positive case.  If you have an implementation of your idea to test that would be good.

jiegillet: OK I did: [here](https://github.com/jiegillet/elixir-analyzer/commit/90182774492c4932f60c5aa0245ffe72b717e5f0) and [here](https://github.com/jiegillet/elixir-analyzer/commit/7c6156caccbcf3c009c1f596108d3d4470225bec). 
(sorry my GitHub-Foo is a bit weak, there is probably a better way to link to these)

angelikatyborska: I took over @jiegillet's test, solution, and added a bunch more tests for more corner cases. Thank  you!

jiegillet: Neat, tabs and escaped tabs into multiline strings became very strange to me after a while, the solution was obvious: add more tests :)

neenjaw (approved):

Issue #89, opened 2021-07-07
Write analysis for concept exercise `need-for-speed`


angelikatyborska:
Desired checks: https://github.com/exercism/elixir/blob/9f1dbea819560a8456b34e03c7819f3d87d11140/exercises/concept/need-for-speed/.meta/design.md#L27-L31

Issue #93, opened 2021-07-07
Write analysis for concept exercise `rpn-calculator-output`


angelikatyborska:
Desired checks: https://github.com/exercism/elixir/blob/9774308bb6b898df6ecbde64c5c6e48c651870ae/exercises/concept/rpn-calculator-output/.meta/design.md#L25-L31

jiegillet:
I've been looking at this one, but I got stuck somewhere.

Two of the criteria (there are more, I'm simplifying) are
* open must be before try
* write must be in the try block

For example:
```elixir
{:ok, file} = resource.open(filename)
try do
  IO.write(file, equation)
rescue 
  _ -> :error
end
```
This is OK. A naive implementation with `feature` would be
```elixir
feature "open must be before try" do
  form do
    _ignore = resource.open(filename)
    try _ignore
  end
end
```
plus 
```elixir
feature "write must be in the try block" do
  form do
    try do: IO.write(_ignore, _ignore), rescue: _ignore
  end
end
```
However, when I try this failing solution
```elixir
try do
  {:ok, file} = resource.open(filename)
  IO.write(file, equation)
rescue 
  _ -> :error
end
```
both of my features fail, because `"write must be in the try block"` only pattern matches a single line. Of course I could add one feature with an `_ignore` one top of `write`, another with one below, but that doesn't feel right.

Is there a feature of `feature` that I'm missing? 
I'm envisioning something like 
```elixir
feature "write must be in the try block" do
  form do
    try _block_includes
      IO.write(_ignore, _ignore)
   rescue
     _ignore
  end
end
```
I haven't fully grasped the code yet, so I'm not sure if it's doable. Please advise :)

angelikatyborska:
> Of course I could add one feature with an _ignore one top of write, another with one below, but that doesn't feel right.

> Is there a feature of feature that I'm missing?

No, I'm afraid that's the "beauty" of `feature`. I often have problems with using it precisely for the same reason.

It's possible that the desired checks that we described aren't even possible with the current features of the analyzer.

The `assert_call` and `feature` macros are really nice for some checks, but in some cases I just wish we could also write an analyzer extension for an exercise by pattern matching on the AST directly. Then anything is possible! @neenjaw what do you think about this problem?

neenjaw:
Considering feature does pattern match against the AST, I don't think what you are describing is really a feature use-case. 

[...]

PR #118, opened 2021-07-27
[#89] Write analysis for concept exercise `need-for-speed`

jiegillet:
I hit a wall with this one and I need some help.

I think I found a bug related to `feature` matching multiline strings and string interpolation... Maybe?

I added some tests in `test/elixir_analyzer/exercise_test/feature/string_interpolation_test.exs` the first 3 pass and the next ones fail. I'm not sure where to begin.

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -51,6 +51,9 @@ config :elixir_analyzer,
     "take-a-number" => %{
       analyzer_module: ElixirAnalyzer.TestSuite.TakeANumber
     },
+    "need-for-speed" => %{
+      analyzer_module: ElixirAnalyzer.TestSuite.NeedForSpeed
+    },
 
     # practice exercises
     "accumulate" => %{
--- a/lib/elixir_analyzer/constants.ex
+++ b/lib/elixir_analyzer/constants.ex
@@ -61,6 +61,11 @@ defmodule ElixirAnalyzer.Constants do
     # Take A Number Comments
     take_a_number_do_not_use_abstractions: "elixir.take-a-number.do_not_use_abstractions",
 
+    # Need For Speed Comments
+    need_for_speed_import_IO_with_only: "elixir.need-for-speed.import_IO_with_only",
+    need_for_speed_import_ANSI_with_except: "elixir.need-for-speed.import_ANSI_with_except",
+    need_for_speed_do_not_modify_code: "elixir.need-for-speed.do_not_modify_code",
+
     # Practice exercises
 
     # Accumulate Comments
--- a/lib/elixir_analyzer/exercise_test.ex
+++ b/lib/elixir_analyzer/exercise_test.ex
@@ -44,17 +44,18 @@ defmodule ElixirAnalyzer.ExerciseTest do
       def analyze(%Submission{} = submission, code_as_string) when is_binary(code_as_string) do
         case Code.string_to_quoted(code_as_string) do
           {:ok, code_ast} ->
-            analyze(submission, code_ast)
+            do_analyze(submission, code_ast, code_as_string)
 
           {:error, e} ->
             append_analysis_failure(submission, e)
         end
       end
 
-      def analyze(%Submission{} = submission, code_ast) do
+      defp do_analyze(%Submission{} = submission, code_ast, code_as_string)
+           when is_binary(code_as_string) do
         feature_results = unquote(feature_tests) |> filter_suppressed_results()
         assert_call_results = unquote(assert_call_tests) |> filter_suppressed_results()
-        common_checks_results = CommonChecks.run(code_ast, Macro.to_string(code_ast))
+        common_checks_results = CommonChecks.run(code_ast, code_as_string)
 
         submission
         |> append_test_comments(feature_results)
--- a/lib/elixir_analyzer/exercise_test.ex
+++ b/lib/elixir_analyzer/exercise_test.ex
@@ -40,22 +40,26 @@ defmodule ElixirAnalyzer.ExerciseTest do
     assert_call_tests = Enum.map(assert_call_data, &AssertCallCompiler.compile(&1, code_ast))
 
     quote do
-      @spec analyze(Submission.t(), String.t()) :: Submission.t()
-      def analyze(%Submission{} = submission, code_as_string) do
+      @spec analyze(Submission.t(), String.t() | any()) :: Submission.t()
+      def analyze(%Submission{} = submission, code_as_string) when is_binary(code_as_string) do
         case Code.string_to_quoted(code_as_string) do
           {:ok, code_ast} ->
-            feature_results = unquote(feature_tests) |> filter_suppressed_results()
-            assert_call_results = unquote(assert_call_tests) |> filter_suppressed_results()
-            common_checks_results = CommonChecks.run(code_ast, code_as_string)
-
-            submission
-            |> append_test_comments(feature_results)
-            |> append_test_comments(assert_call_results)
-            |> append_test_comments(common_checks_results)
+            analyze(submission, code_ast)
 
           {:error, e} ->
             append_analysis_failure(submission, e)
         end
+      end
+
+      def analyze(%Submission{} = submission, code_ast) do
+        feature_results = unquote(feature_tests) |> filter_suppressed_results()
+        assert_call_results = unquote(assert_call_tests) |> filter_suppressed_results()
+        common_checks_results = CommonChecks.run(code_ast, Macro.to_string(code_ast))
+
+        submission
+        |> append_test_comments(feature_results)
+        |> append_test_comments(assert_call_results)
+        |> append_test_comments(common_checks_results)
       end
 
       defp filter_suppressed_results(feature_results) do
--- a/lib/elixir_analyzer/exercise_test.ex
+++ b/lib/elixir_analyzer/exercise_test.ex
@@ -40,7 +40,7 @@ defmodule ElixirAnalyzer.ExerciseTest do
     assert_call_tests = Enum.map(assert_call_data, &AssertCallCompiler.compile(&1, code_ast))
 
     quote do
-      @spec analyze(Submission.t(), String.t() | any()) :: Submission.t()
+      @spec analyze(Submission.t(), String.t()) :: Submission.t()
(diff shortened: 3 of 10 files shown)


angelikatyborska:
>  I'm not sure where to begin.

Those are general hints that could make debugging easier but I don't know really what's going on:

I would start by making your intentions more clear by specifying a `find` value in the `feature` calls. 

Then, for every test that you want to pass, I would add an opposite test. This is to make sure that the check actually works and isn't ignored. E.g.

for
```elixir
  test_exercise_analysis "normal string",
    comments_exclude: ["normal string"] do
    [
      defmodule MyModule do
        def hi(name) do
          "hello you"
        end
      end
    ]
  end
```
add
```elixir
  test_exercise_analysis "not a normal string",
    comments_include: ["normal string"] do
    [
      defmodule MyModule do
        def hi(name) do
          "hello you!!!!!!"
        end
      end
    ]
  end
```

Then, I would comment out all the failing tests but the first one and their corresponding `feature` calls so that you can focus on a single weird case.

Another tip, it's impossible to tell apart multiline strings from normal strings by the AST so I would drop that distinction (or make it clear in the tests that you know it's the same thing):

```
iex(4)> quote do
...(4)> """
...(4)> hello
...(4)> """
...(4)> end
"hello\n"

iex(5)> quote do
...(5)> "hello\n"
...(5)> end
"hello\n"
```


I think it is possible that the `feature` macro is unable to handle blocks that are _just_ strings with interpolation and no other calls 🤔 but I don't know why atm...

neenjaw:
Had a spot of internet, I think it is because `"""` is a macro in and of itself.  The current match algorithm depend on the quoted ast of the form to expand all at once, so that we can replace things (lib/elixir_analyzer/exercise_test/feature.ex Line 75-122) and create a pattern that doesn't expand any further when used (lib/elixir_analyzer/exercise_test/feature/compiler.ex Line 48-65).

At least in the case of string interpolation in the context of a multiline string, this seems to fail:

```
iex(2)> quote do  
...(2)>   """     
...(2)>   #{value}\n
...(2)>   """       
...(2)> end         
{:<<>>, [],
 [
   "",
   {:"::", [],
    [
      {{:., [], [Kernel, :to_string]}, [], [{:value, [], Elixir}]},
      {:binary, [], Elixir}
    ]},
   "\n\n"
 ]}
```

This is the quoted form of the code segment.

```
"{:<<>>, _, [\"hello \", {:\"::\", _, [{{:., _, [Kernel, :to_string]}, _, [{:name, _, nil}]}, {:binary, _, nil}]}]}"
```
This is the string pattern created by the feature macro on `lib/elixir_analyzer/exercise_test/feature.ex:L115`

```
{:{}, [line: 1],
 [
   :<<>>,
   {:_, [line: 1], nil},
   [
     "hello ",
     {:{}, [line: 1],
      [
        :"::",
        {:_, [line: 1], nil},
        [
          {:{}, [line: 1],
           [
             {:{}, [line: 1],
              [
                :.,
                {:_, [line: 1], nil},
                [{:__aliases__, [line: 1], [:Kernel]}, :to_string]
              ]},
             {:_, [line: 1], nil},
             [{:{}, [line: 1], [:name, {:_, [line: 1], nil}, nil]}]
           ]},
          {:{}, [line: 1], [:binary, {:_, [line: 1], nil}, nil]}
        ]
      ]},
     "\n"
   ]
 ]}
```
This is the evaluated string created by the feature compiler on `lib/elixir_analyzer/exercise_test/feature/compiler.ex:L52`. Notice the metadata reinserted into the ast.  This will never match anything because of that addition.

I don't know of any Elixir examples of creating custom pattern matchers. And as far as I know it is otherwise impossible to create dynamic patterns to leverage `=/2` or `match?/2`.  Interestingly, ETS might be an example we could follow to create out own matcher.  See this SO post by rvirding: https://stackoverflow.com/questions/3989271/dynamic-pattern-matching


---

TLDR: Right now we can't do this check, but we can pull an ETS and build a custom matcher... maybe

jiegillet:
I see. I saw that pattern when I was debugging around, but did not know that it wasn't supposed to be that way. I guess I'll clean up the tests and keep familiarizing myself with the codebase and stuff :)

> Had a spot of internet

I hope I'm not ruining your vacation :D

neenjaw:
> I hope I'm not ruining your vacation

Nope, just thought it worthwhile to share what I found about why it wasn't working to save some time and headache.

angelikatyborska:
@jiegillet are you stuck? do you need more help?

jiegillet:
I was stuck for a while, so I took a break but kept thinking about it. I've started looking at it again yesterday, and I'm making progress understanding the issue. I think it is a bit different from what we discussed before, but I need a bit more time to be sure :)

jiegillet:
OK, I finally came up with something. This is possibly a big change so I will justify it a bit and I would love to have your input.
Here what I think was going wrong

```elixir
iex(85)> quote do
...(85)>   def hi(name) do
...(85)>     "hello #{name}\n"
...(85)>   end
...(85)> end
{:def, [context: Elixir, import: Kernel],
 [
   {:hi, [context: Elixir], [{:name, [], Elixir}]},
   [
     do: {:<<>>, [],
      [
        "hello ",
        {:"::", [],
         [
           {{:., [], [Kernel, :to_string]}, [], [{:name, [], Elixir}]},
           {:binary, [], Elixir}
         ]},
        "\n"            # Note this single newline...
      ]}
   ]
 ]}
iex(86)> |> Macro.to_string
"def(hi(name)) do\n  \"hello \#{name}\n  \"\nend"
iex(87)> |> tap(&IO.puts/1)
def(hi(name)) do
  "hello #{name}
  "                      # ... that becomes an actual newline instead of being escaped
end
"def(hi(name)) do\n  \"hello \#{name}\n  \"\nend"
iex(88)> |> Code.string_to_quoted!
{:def, [line: 1],
 [
   {:hi, [line: 1], [{:name, [line: 1], nil}]},
   [
     do: {:<<>>, [line: 2],
      [
        "hello ",
        {:"::", [line: 2],
         [
           {{:., [line: 2], [Kernel, :to_string]}, [line: 2],
            [{:name, [line: 2], nil}]},
           {:binary, [line: 2], nil}
         ]},
        "\n  "     # and drags the indentation with it
      ]}
   ]
 ]}
```

Basically `Macro.to_string` confuses the ending newline in the string for an actual newline, write the ending quote indented and the indentation gets dragged in by `Code.string_to_quoted!`. This may be something similar to [this issue](https://github.com/elixir-lang/elixir/issues/5318)? 

So I decided to try not going through the intermediate string and match the AST directly, by replacing `match?` by a custom `form_match?`. It seems to work fine. Some side effects I could think of:
* Good side effect: I think this can solve #93 by letting people use `form` within `form` and using `form_match?` deal with it
* Bad side effect: I haven't checked yet, but this probably breaks #117

angelikatyborska:
🤯 Great detective work!

This is definitely a bug in Elixir. I downloaded 1.12.2 to make sure it's still there, it is. I think we should open an issue in the Elixir GitHub repo. Do you want to do the honors, or should I? I looked through `Macro.to_string` related issues but there are no open issues for this specific problem.

So basically for this input:

```
iex()> quoted =
...()> quote do
...()>   def hi(name) do
...()>     "hello #{name}\n"
...()>   end
...()> end
{:def, [context: Elixir, import: Kernel],
 [
   {:hi, [context: Elixir], [{:name, [], Elixir}]},
   [
     do: {:<<>>, [],
      [
        "hello ",
        {:"::", [],
         [
           {{:., [], [Kernel, :to_string]}, [], [{:name, [], Elixir}]},
           {:binary, [], Elixir}
         ]},
        "\n"
      ]}
   ]
 ]}

iex()> string = Macro.to_string(quoted)
```

Actual output
```elixir
"def(hi(name)) do\n  \"hello \#{name}\n  \"\nend"
```
```elixir
# IO.puts(string)
def(hi(name)) do
  "hello #{name}
  "
end
```

Expected output
```elixir
"def(hi(name)) do\n  \"hello \#{name}\\n\"\nend"
```
```elixir
def(hi(name)) do
  "hello #{name}\n"
end
```

> Bad side effect: I haven't checked yet, but this probably breaks

It does 🤔 I'll look into it, maybe it's possible to fix?

angelikatyborska:
@jiegillet Here's my proposal on how to fix #117 and keep your changes, in a commit: https://github.com/exercism/elixir-analyzer/commit/b27e59d4d990e1a402723e385904a5519ea4b9e2

The real input to the analyzer will be a string. So for tests where this weird bug changes the behavior, I propose we also define the test data as a string.

jiegillet:
> This is definitely a bug in Elixir. I downloaded 1.12.2 to make sure it's still there, it is. I think we should open an issue in the Elixir GitHub repo. Do you want to do the honors, or should I? I looked through `Macro.to_string` related issues but there are no open issues for this specific problem.

Alright, [done](https://github.com/elixir-lang/elixir/issues/11172). Thank you for your help!

> The real input to the analyzer will be a string. So for tests where this weird bug changes the behavior, I propose we also define the test data as a string.

Sure, that makes sense. I'm glad you were able to align #117 so quickly.
What's the next step? Reviews from @neenjaw? 
I don't really want to write new analyses if the AST matching is going to change.

angelikatyborska:
From your side, you could merge main to this branch to resolve conflicts, and then you can cherry-pick my fix into this branch too (https://github.com/exercism/elixir-analyzer/commit/b27e59d4d990e1a402723e385904a5519ea4b9e2) because the changes are all for code from this branch, not from #117. I think there are dialyzer warnings, if that's true they might be unsolvable and might need to be ignored, e.g. like here https://github.com/exercism/elixir-analyzer/blob/b27e59d4d990e1a402723e385904a5519ea4b9e2/lib/elixir_analyzer/exercise_test.ex#L20

I'll ping Tim to see if he's back and interested in a review. If not, I'll approve and merge 🙂

jiegillet:
Well, the issue is already closed
> This has been fixed in Elixir master since Macro.to_string has been rewritten to use the same code path as the code formatter. Thank you!

angelikatyborska (approved):

angelikatyborska:
Tim says we can merge without his review 🙂

PR #119, opened 2021-08-04
Fix variable name check not reporting when using guards

angelikatyborska:


test/elixir_analyzer/exercise_test/common_checks/function_names_test.exs
```diff
@@ -40,8 +40,8 @@ defmodule ElixirAnalyzer.ExerciseTest.CommonChecks.FunctionNamesTest do
           defp registered_in_last_quarter?(user), do: true
           defmacro validate_user!, do: true
           defmacrop do_validate_user, do: true
-          defguard(is_user, do: true)
-          defguardp(is_registered_user, do: true)
+          defguard is_user when true
+          defguardp is_registered_user when true
```
angelikatyborska, line 44: If I added semantically correct guard definitions in the first place, I would have caught the issue 🤦

jiegillet (approved):
Looks good to me. Nice catch!

neenjaw (approved):

PR #123, opened 2021-08-17
[#91] Write analysis for concept exercise `rpg-character-sheet`

### LINKED ISSUE #91: Write analysis for concept exercise `rpg-character-sheet`
Desired checks: https://github.com/exercism/elixir/blob/9774308bb6b898df6ecbde64c5c6e48c651870ae/exercises/concept/rpg-character-sheet/.meta/design.md#L24-L29

jiegillet:
Closes #91. Fixes #91. Does not solve #91.

This PR introduces `_block_ends_with`, a block very similar to `_block_includes`, but the last line must match the last line of the code. This is useful for checking the return value of functions, like in `rpg-character-sheet`.

When trying to match a single line, one must be careful to always put in in a context (module definition, function definition,...) otherwise the match is done line to line and can succeed even when the line isn't the last in the code. I explain that in the docs.

I copy/pasted and modified the tests for `_block_includes`, I'm not sure if that's good practice or if I should have tried to merge the two. Please let me know.

### DIFF
--- a/docs/recipes.md
+++ b/docs/recipes.md
@@ -95,3 +95,22 @@ feature "description" do
   end
 end
 ```
+
+## Asserting that a block finished with a specific function 
+
+This check will also pass if there are other function calls in between or before (bot not after). When used to match a single line, make sure to place `_block_ends_with` inside a context (like a module or function definition) as in the example below.
+
+```elixir
+feature "description" do
+  find :any
+
+  form do
+    def read_file(_ignore) do
+      _block_ends_with do
+        _ignore = File.open(_ignore)
+        File.close(_ignore)
+      end
+    end
+  end
+end
+```
--- a/docs/recipes.md
+++ b/docs/recipes.md
@@ -98,7 +98,7 @@ end
 
 ## Asserting that a block finished with a specific function 
 
-This check will also pass if there are other function calls in between or before (bot not after). When used to match a single line, make sure to place `_block_ends_with` inside a context (like a module or function definition) as in the example below.
+This check will also pass if there are other function calls in between or before (but not after). When used to match a single line, make sure to place `_block_ends_with` inside a context (like a module or function definition) as in the example below.
 
 ```elixir
 feature "description" do
--- a/lib/elixir_analyzer/test_suite/rpg_character_sheet.ex
+++ b/lib/elixir_analyzer/test_suite/rpg_character_sheet.ex
@@ -1,0 +1,89 @@
+defmodule ElixirAnalyzer.TestSuite.RpgCharacterSheet do
+  @moduledoc """
+  This is an exercise analyzer extension module for the concept exercise rpg-character-sheet
+  """
+
+  use ElixirAnalyzer.ExerciseTest
+  alias ElixirAnalyzer.Constants
+
+  feature "welcome ends with IO.puts" do
+    type :essential
+    find :all
+    depth 1
+    comment Constants.rpg_character_sheet_welcome_ends_with_IO_puts()
+
+    form do
+      def welcome() do
+        _block_ends_with do
+          IO.puts(_ignore)
+        end
+      end
+    end
+  end
+
+  assert_call "run uses welcome" do
+    type :essential
+    called_fn name: :welcome
+    calling_fn module: RPG.CharacterSheet, name: :run
+    comment ElixirAnalyzer.Constants.rpg_character_sheet_run_uses_other_functions()
+  end
+
+  assert_call "run uses ask_name" do
+    type :essential
+    called_fn name: :ask_name
+    calling_fn module: RPG.CharacterSheet, name: :run
+    comment ElixirAnalyzer.Constants.rpg_character_sheet_run_uses_other_functions()
+  end
+
+  assert_call "run uses ask_class" do
+    type :essential
+    called_fn name: :ask_class
+    calling_fn module: RPG.CharacterSheet, name: :run
+    comment ElixirAnalyzer.Constants.rpg_character_sheet_run_uses_other_functions()
+  end
+
+  assert_call "run uses ask_level" do
+    type :essential
+    called_fn name: :ask_level
+    calling_fn module: RPG.CharacterSheet, name: :run
+    comment ElixirAnalyzer.Constants.rpg_character_sheet_run_uses_other_functions()
+  end
+
+  feature "run ends with IO.inspect" do
+    type :essential
+    find :one
+    depth 1
+    comment Constants.rpg_character_sheet_run_ends_with_IO_inspect()
+
+    form do
+      def run() do
+        _block_ends_with do
+          IO.inspect(_ignore)
+        end
+      end
+    end
+
+    form do
+      def run() do
+        _block_ends_with do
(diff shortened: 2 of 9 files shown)


jiegillet:
I set all the features as `:essential`, but after writing the descriptions, I feel that it's a bit too much. Maybe the two `IO.inspect` warrant that, because learning how that function works is super useful, but the other two feel more like `:actionable`. What do you think?

lib/elixir_analyzer/test_suite/rpg_character_sheet.ex
```diff
@@ -0,0 +1,89 @@
...
+    calling_fn module: RPG.CharacterSheet, name: :run
+    comment ElixirAnalyzer.Constants.rpg_character_sheet_run_uses_other_functions()
+  end
+
+  assert_call "run uses ask_level" do
+    type :essential
+    called_fn name: :ask_level
+    calling_fn module: RPG.CharacterSheet, name: :run
+    comment ElixirAnalyzer.Constants.rpg_character_sheet_run_uses_other_functions()
+  end
```
angelikatyborska, line 50: > Maybe the two IO.inspect warrant that, because learning how that function works is super useful, but the other two feel more like :actionable. What do you think?

I agree with your reasoning and conclusions. That would mean this change, right?

```suggestion
  feature "welcome ends with IO.puts" do
    type :actionable
    find :all
    depth 1
    comment Constants.rpg_character_sheet_welcome_ends_with_IO_puts()

    form do
      def welcome() do
        _block_ends_with do
          IO.puts(_ignore)
        end
      end
    end
  end

  assert_call "run uses welcome" do
    type :actionable
    called_fn name: :welcome
    calling_fn module: RPG.CharacterSheet, name: :run
    comment ElixirAnalyzer.Constants.rpg_character_sheet_run_uses_other_functions()
  end

  assert_call "run uses ask_name" do
    type :actionable
    called_fn name: :ask_name
    calling_fn module: RPG.CharacterSheet, name: :run
    comment ElixirAnalyzer.Constants.rpg_character_sheet_run_uses_other_functions()
  end

  assert_call "run uses ask_class" do
    type :actionable
    called_fn name: :ask_class
    calling_fn module: RPG.CharacterSheet, name: :run
    comment ElixirAnalyzer.Constants.rpg_character_sheet_run_uses_other_functions()
  end

  assert_call "run uses ask_level" do
    type :actionable
    called_fn name: :ask_level
    calling_fn module: RPG.CharacterSheet, name: :run
    comment ElixirAnalyzer.Constants.rpg_character_sheet_run_uses_other_functions()
  end
```

test/elixir_analyzer/exercise_test/feature/block_ends_with_test.exs
```diff
@@ -0,0 +1,477 @@
...
+            :ok.lets_pretend_this_is_an_erlang_module()
+            # Erlang function call counts as atom literal
+          end
+        end,
+        defmodule MyModule do
+          def foo() do
+            :ok
+            # Atoms are always found, even when they are not the last
+            :nope
+          end
```
angelikatyborska, line 48: What's going on here? Why does it find the `:ok` atom even though it's not last?

test/elixir_analyzer/exercise_test/feature/block_ends_with_test.exs
```diff
@@ -0,0 +1,477 @@
...
+        def baz(_, _), do: :ok
+
+        def foo(), do: :hello
+
+        def bar(baz), do: :bye
+      end
+    ]
+  end
+
+  test_exercise_analysis "code not endinf with functions",
```
angelikatyborska, line 344: Typo
```suggestion
  test_exercise_analysis "code not ending with functions",
```

test/elixir_analyzer/test_suite/rpg_character_sheet_test.exs
```diff
@@ -0,0 +1,234 @@
...
+      end,
+      defmodule RPG.CharacterSheet do
+        def welcome() do
+          text = "Welcome! Let's fill out your character sheet together."
+          IO.puts(text)
+          :ok
+        end
+      end
+    ]
+  end
```
angelikatyborska, line 64: Since this check is about the implicit vs explicit `:ok` return, I would like to allow students to use `IO.write` instead of `IO.puts`. Could we add a test and make it pass?

The (already merged) comment can stay like it is. We don't need to customize if to say "puts or write" because it _should_ be puts, so maybe the student would get that hint from the comment too 🙂 

```suggestion
  describe "implicit :ok return" do
    test_exercise_analysis "ending with IO.write is not ideal, but allowed",
      comments_exclude: [Constants.rpg_character_sheet_welcome_ends_with_IO_puts()] do
      [
        defmodule RPG.CharacterSheet do
          def welcome() do
            IO.write("Welcome! Let's fill out your character sheet together.\n")
          end
        end
      ]
    end

    test_exercise_analysis "welcome doesn't end with IO.puts or IO write",
      comments_include: [Constants.rpg_character_sheet_welcome_ends_with_IO_puts()] do
      [
        defmodule RPG.CharacterSheet do
          def welcome() do
            IO.puts("Welcome! Let's fill out your character sheet together.")
            :ok
          end
        end,
        defmodule RPG.CharacterSheet do
          def welcome() do
            text = "Welcome! Let's fill out your character sheet together."
            IO.puts(text)
            :ok
          end
        end,
        defmodule RPG.CharacterSheet do
          def welcome() do
            text = "Welcome! Let's fill out your character sheet together."
            IO.write(text <> "\n")
            :ok
          end
        end
      ]
    end
  end
```

docs/recipes.md
```diff
@@ -95,3 +95,22 @@ feature "description" do
   end
 end
 ```
+
+## Asserting that a block finished with a specific function 
+
+This check will also pass if there are other function calls in between or before (bot not after). When used to match a single line, make sure to place `_block_ends_with` inside a context (like a module or function definition) as in the example below.
```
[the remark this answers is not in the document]
jiegillet: Hahaha, maybe I meant bots cannot come after either 🤖🤖🤖

test/elixir_analyzer/exercise_test/feature/block_ends_with_test.exs
jiegillet: Maybe the comment is not accurate. With this form:

```elixir 
form do
  _block_ends_with do
    :ok
  end
end
```
`_block_ends_with` will get compared to every single node in the code AST, not just groups of node, so it will match any `:ok` anywhere (because a single node is also technically the last).

However here
```elixir 
form do
  def foo() do
    _block_ends_with do
      :ok
    end
  end
end
```
`_block_ends_with` will only get compared to the group of lines inside of `foo`, so it will only match an `:ok` that is at the last line. I've added some tests that make this behavior ore clear.

I know this is confusing, but I can't really think of a way out here, other than writing clear documentation.

angelikatyborska: Ah, I get it now. The updated tests are enough, thanks 🙂

angelikatyborska (approved):

PR #124, opened 2021-08-19
[#25] Raise error if forms within a feature are not unique

### LINKED ISSUE #25: Raise error if forms within a feature are not unique
Certain differences in syntax cannot be told apart from one another by the AST, for example:

```elixir
a =
  quote do
    def foo do
      "hello"
    end
  end

b =
  quote do
    def foo, do: "hello"
  end

a == b
# => true
```

This might not be known to everyone that attempts writing exercise test modules. It would be helpful to automatically detect attempts to specify multiple `form`s within the same `feature` that are identical, to educate and to avoid unnecessarily repetitive code.

See: https://github.com/exercism/elixir-analyzer/pull/24/files#r512784658

jiegillet:
Closes #25.

This one was very fun, I really like the structure of the tests ([found here](https://elixirforum.com/t/how-to-test-a-compile-error/15626)).

I am testing for 2 things: if forms within a feature are the same, and if features within a test module are the same.
I even found a duplicate feature in our current set :)

### DIFF
--- a/lib/elixir_analyzer/exercise_test/feature.ex
+++ b/lib/elixir_analyzer/exercise_test/feature.ex
@@ -4,8 +4,7 @@ defmodule ElixirAnalyzer.ExerciseTest.Feature do
   whose AST matches part of the AST of the solution.
   """
 
-  alias ElixirAnalyzer.ExerciseTest.Feature
-  defexception message: "A duplicate feature has been found"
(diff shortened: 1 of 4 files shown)


angelikatyborska:
Awesome!

> I even found a duplicate feature in our current set :)

I can't say I'm surprised 😇

I'll wait a few days with merging to see if @neenjaw wants to add something.

angelikatyborska (approved):

PR #126, opened 2021-08-22
[#40] Write analysis for practice exercise `list-ops`

### LINKED ISSUE #40: Write a basic `list-ops` analysis
There are many standard library functions that are not allowed when solving this exercise. See comment:
https://github.com/exercism/elixir/blob/main/exercises/practice/list-ops/lib/list_ops.ex#L2-L6

Please implement an analyzer that will ensure those not allowed functions aren't used.

Introduction to writing an analyzer extension: https://github.com/exercism/elixir-analyzer/blob/main/docs/writing-an-analyzer.md

An example of an analyzer extension that detect usages of disallowed modules: https://github.com/exercism/elixir-analyzer/pull/50/files#diff-39ba60165e48ed55ff5dc41ce966786ae8f1c83abf1cc3e2430e0b8b5692f105

jiegillet:
Fixes #40.

I could not find a reasonable way of using `in` in an implementation so I left that test out.
I added `for` to the list of forbidden functions.

On the way, I had to patch the `assert_call` compiler because it would not flag a function names used in a function definition of the same name so `def length(l), do: Kernel.length(l)` would not get flagged. This lead me to find bugs in some `high-score` tests.

### DIFF
--- a/lib/elixir_analyzer/exercise_test/assert_call/compiler.ex
+++ b/lib/elixir_analyzer/exercise_test/assert_call/compiler.ex
@@ -51,6 +51,7 @@ defmodule ElixirAnalyzer.ExerciseTest.AssertCall.Compiler do
 
   def assert(ast, called_fn, calling_fn) do
     acc = %{
+      in_module: nil,
       in_function_def: nil,
       in_function_modules: %{},
       modules_in_scope: %{},
@@ -81,15 +82,10 @@ defmodule ElixirAnalyzer.ExerciseTest.AssertCall.Compiler do
       |> track_aliases(node)
       |> track_imports(node)
 
-    function_def? = function_def?(node)
-    name = extract_function_name(node)
-
-    case {function_def?, name} do
-      {true, name} ->
-        {node, %{acc | in_function_def: name}}
-
-      _ ->
-        {node, acc}
+    cond do
+      module_def?(node) -> {node, %{acc | in_module: extract_module_name(node)}}
+      function_def?(node) -> {node, %{acc | in_function_def: extract_function_name(node)}}
+      true -> {node, acc}
     end
   end
 
@@ -100,10 +96,10 @@ defmodule ElixirAnalyzer.ExerciseTest.AssertCall.Compiler do
   def annotate_and_find(node, acc) do
     {node, acc} = find(node, acc)
 
-    if function_def?(node) do
-      {node, %{acc | in_function_def: nil, in_function_modules: %{}}}
-    else
-      {node, acc}
+    cond do
+      module_def?(node) -> {node, %{acc | in_module: nil}}
+      function_def?(node) -> {node, %{acc | in_function_def: nil, in_function_modules: %{}}}
+      true -> {node, acc}
     end
   end
 
@@ -116,6 +112,7 @@ defmodule ElixirAnalyzer.ExerciseTest.AssertCall.Compiler do
   def find(
         node,
         %{
+          in_module: module,
           modules_in_scope: modules_in_scope,
           in_function_modules: in_function_modules,
           called_fn: called_fn,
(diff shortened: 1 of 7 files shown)


lib/elixir_analyzer/exercise_test/assert_call/compiler.ex
```diff
@@ -81,15 +82,10 @@ defmodule ElixirAnalyzer.ExerciseTest.AssertCall.Compiler do
...
-    case {function_def?, name} do
-      {true, name} ->
-        {node, %{acc | in_function_def: name}}
-
-      _ ->
-        {node, acc}
+    cond do
+      module_def?(node) -> {node, %{acc | in_module: extract_module_name(node)}}
+      function_def?(node) -> {node, %{acc | in_function_def: extract_function_name(node)}}
+      true -> {node, acc}
```
jiegillet, line 88: In Haskell, the keyword `otherwise` is an alias for `True`, it makes this kind of check a lot more readable. I always find awkward writing this
```elixir
cond do
  this?(x) -> also_that?(x)
  that?(x) -> also_this?(x)
  true -> true
end
```
I think this is much smoother:
```elixir
cond do
  this?(x) -> also_that?(x)
  that?(x) -> also_this?(x)
  otherwise -> true
end
```
Do you think that's a valid feature request? :)

test/elixir_analyzer/test_suite/high_score_test.exs
```diff
@@ -34,7 +34,7 @@ defmodule ElixirAnalyzer.ExerciseTest.HighScoreTest do
   test_exercise_analysis "requires add_player to have a default argument that's a module attribute",
     comments_include: [Constants.high_score_use_default_argument_with_module_attribute()] do
     [
-      def HighScore do
+      defmodule HighScore do
```
angelikatyborska, line 37: Oops 😂

lib/elixir_analyzer/exercise_test/assert_call/compiler.ex
angelikatyborska: Haha 😀. I'm not sure about Haskell, but in Elixir `true` is just syntactic sugar for the atom `:true`. You can already write:

```elixir
cond do
  this?(x) -> also_that?(x)
  that?(x) -> also_this?(x)
  :otherwise -> true
end
```

Because all atoms are truthy and `cond` doesn't require a strictly boolean expression.

There are only three atoms that have this extra treatment that you can write them without `:` - `true`, `false`, and `nil`. I'm not sure if it's such a pressing problem that they would want to add `otherwise` to that list 😅. But I do see the problems. I hate conds that look like this:

```elixir
cond do
  something == false -> :ok
  true -> :error
end
```

The pairing of `false` and `true` here makes my brain hurt.

angelikatyborska (approved):
> I could not find a reasonable way of using in in an implementation so I left that test out.

I can't either 🤷 

>  I added for to the list of forbidden functions.

Good call!

> On the way, I had to patch the assert_call compiler because it would not flag a function names used in a function definition of the same name so def length(l), do: Kernel.length(l) would not get flagged.

👍 I pushed a small test for `assert_no_call` just in case we ever drop the analysis of `list-ops` for any reason, so that this corner case is still covered by the tests

lib/elixir_analyzer/exercise_test/assert_call/compiler.ex
jiegillet: I would not call that a pressing problem, no, but maybe a nice little quality of life update, like Elixir does :)

`:otherwise -> true` is... almost there, but not quite. It might appease me, but definitely not most people who would need to use my code after me!

I share your brain pain!

jiegillet:
> 👍 I pushed a small test for `assert_no_call` just in case we ever drop the analysis of `list-ops` for any reason, so that this corner case is still covered by the tests

👍 
Thank you, I should have done it myself 🙈

PR #129, opened 2021-08-29
[#90] Write analysis for concept exercise `newsletter`

### LINKED ISSUE #90: Write analysis for concept exercise `newsletter`
Desired checks:
https://github.com/exercism/elixir/blob/9774308bb6b898df6ecbde64c5c6e48c651870ae/exercises/concept/newsletter/.meta/design.md#L25-L28

jiegillet:
Closes #90.

Disclaimer: @angelikatyborska I know this is a terribly busy time for you, this can totally wait :)

This PR is not finished, I have some questions first. The instructions in `design.md` say:
- ensure that `send_newsletter/3` calls `open_log/1` and `close_log/1` exactly once each, and doesn't use `File.write`.

I tried to make examples of code with using `open_log` more than once, but it felt silly (why write that command twice in a row?) unless they used `open_log` inside of a loop for each email. However to pass the tests in tat situation, I had to change 
```elixir
def open_log(path) do
  File.open!(path, [:write])
end
```
to
```elixir
def open_log(path) do
  File.open!(path, [:append])
end
```
So I thought that I would replace that test by "open_log uses the :write option" might make more sense, which is what is committed now.  Then, thinking more about  it, I realized that situations could very nicely be avoided by adding one more test in the test suite:

```elixir
    @tag task_id: 5
    test "sending the same newsletter twice resets the log" do
      send_fun = fn _ -> :ok end 
      Newsletter.send_newsletter(Path.join(["assets", "emails.txt"]), @temp_file_path, send_fun)
      Newsletter.send_newsletter(Path.join(["assets", "emails.txt"]), @temp_file_path, send_fun)

      assert File.read!(@temp_file_path) ==
               """ 
               alice@example.com
               bob@example.com
               charlie@example.com
               dave@example.com
               """
    end 
```
That being said, maybe a logger _should_ append and not reset.

So I am full of doubts here. 

1. What use case did you want to prevent by using writing "ensure that `send_newsletter/3` calls `open_log/1` and `close_log/1` exactly once each"?
2. Should I make sure they use `:write` (the instructions are pretty clear that they should)?
3. Should I check for both 1 and 2?
3. Should I add "sending the same newsletter twice resets the log" to the test suite? In that case, maybe 2 becomes useless.

### DIFF
--- a/lib/elixir_analyzer/test_suite/newsletter.ex
+++ b/lib/elixir_analyzer/test_suite/newsletter.ex
@@ -47,8 +47,17 @@ defmodule ElixirAnalyzer.TestSuite.Newsletter do
   end
 
   feature "open_log used the option :write" do
+    find :any
     type :essential
     comment Constants.newsletter_open_log_uses_option_write()
+
+    form do
+      def open_log(_ignore) do
+        _block_includes do
+          File.open(_ignore, [:write])
+        end
+      end
+    end
 
     form do
       def open_log(_ignore) do
--- a/lib/elixir_analyzer/test_suite/newsletter.ex
+++ b/lib/elixir_analyzer/test_suite/newsletter.ex
@@ -59,6 +59,34 @@ defmodule ElixirAnalyzer.TestSuite.Newsletter do
     end
   end
 
+  assert_call "send_newsletter/3 calls open_log/1" do
+    type :actionable
+    comment ElixirAnalyzer.Constants.newsletter_send_newsletter_reuses_functions()
+    called_fn name: :open_log
+    calling_fn module: Newsletter, name: :send_newsletter
+  end
+
+  assert_call "send_newsletter/3 calls close_log/1" do
+    type :actionable
+    comment ElixirAnalyzer.Constants.newsletter_send_newsletter_reuses_functions()
+    called_fn name: :close_log
+    calling_fn module: Newsletter, name: :send_newsletter
+  end
+
+  assert_call "send_newsletter/3 calls read_emails/1" do
+    type :actionable
+    comment ElixirAnalyzer.Constants.newsletter_send_newsletter_reuses_functions()
+    called_fn name: :read_emails
+    calling_fn module: Newsletter, name: :send_newsletter
+  end
+
+  assert_call "send_newsletter/3 calls log_sent_email/2" do
+    type :actionable
+    comment ElixirAnalyzer.Constants.newsletter_send_newsletter_reuses_functions()
+    called_fn name: :log_sent_email
+    calling_fn module: Newsletter, name: :send_newsletter
+  end
+
   assert_no_call "send_newsletter/3 doesn't use File.write" do
     type :essential
     comment ElixirAnalyzer.Constants.newsletter_send_newsletter_does_not_call_write()
--- a/lib/elixir_analyzer/test_suite/newsletter.ex
+++ b/lib/elixir_analyzer/test_suite/newsletter.ex
@@ -86,18 +86,4 @@ defmodule ElixirAnalyzer.TestSuite.Newsletter do
     called_fn name: :log_sent_email
     calling_fn module: Newsletter, name: :send_newsletter
   end
-
-  assert_no_call "send_newsletter/3 doesn't use File.write" do
-    type :essential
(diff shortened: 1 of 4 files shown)


angelikatyborska:
> 1. What use case did you want to prevent by using writing "ensure that send_newsletter/3 calls open_log/1 and close_log/1 exactly once each"?

I thought we need to prevent open + close within the loop because I wasn't aware that it wouldn't work with the desired implementation of `open_log` 😅. You're right, this check makes no sense. But maybe it makes sense to check for function reuse in general? E.g. ensure that `send_newsletter` does use `open_log`, `close_log`, and `log_sent_email` at least once.

> 2. Should I make sure they use :write (the instructions are pretty clear that they should)?

> 4. Should I add "sending the same newsletter twice resets the log" to the test suite? In that case, maybe 2 becomes useless.

Yes to both! Adding a unit test is better because the feedback about failure comes before submitting the exercise. But I think that the unit test won't be able to verify everything (e.g. write vs write + read), so an analyzer check also makes sense.

When making those changes to the tests, can you also update the `design.md` for this exercise to reflect the checks?

I'm glad you thought through those requirements from design.md. I definitely didn't put much effort into that when I wrote them 😂

lib/elixir_analyzer/test_suite/newsletter.ex
```diff
@@ -0,0 +1,89 @@
...
+
+  feature "close_log ends with File.close" do
+    type :actionable
+    comment Constants.newsletter_close_log_returns_implicitly()
+
+    form do
+      def close_log(_ignore) do
+        _block_ends_with do
+          File.close(_ignore)
+        end
```
angelikatyborska, line 18: Nice how useful the new feature is 🙂 👏

It will cause a false-positive if somebody uses `IO.inspect` as the last call or assigns the return value to a variable and puts that variable as the last expression. But both are bad style so I don't care. Debug calls to `IO.inspect` should be removed before submitting a solution and the `x = File.close(x); x` is a compilation warning about a redundant variable.

I was thinking about adding a common check for `IO.inspect` in all exercises but a few that actually require it (rpg character sheet I think). To prompt people to clean up the debug calls. What do you think?

lib/elixir_analyzer/test_suite/newsletter.ex
```diff
@@ -0,0 +1,89 @@
...
+  end
+
+  feature "open_log used the option :write" do
+    type :essential
+    comment Constants.newsletter_open_log_uses_option_write()
+
+    form do
+      def open_log(_ignore) do
+        _block_includes do
+          File.open!(_ignore, [:write])
```
angelikatyborska, line 56: There's no requirements and no tests in the exercise about `open/2` vs `open!/2` so I think both should be allowed here.

angelikatyborska (approved):

lib/elixir_analyzer/test_suite/newsletter.ex
jiegillet: > I was thinking about adding a common check for `IO.inspect` in all exercises but a few that actually require it (rpg character sheet I think). To prompt people to clean up the debug calls. What do you think?

Yes, I'm interested in working on that. I'll have to think of a clean way do exclude a common check. First thing that popped into my mind is:
```elixir
defmodule ElixirAnalyzer.TestSuite.RpgCharacterSheet do
  use ElixirAnalyzer.ExerciseTest,
    suppress_check: ElixirAnalyzer.Constants .solution_debug_functions(),
...
end
```
I would then pool all the test (feature/assert_call/common) in `ExerciseTest` and filter out the suppressed ones. It would also allow a `feature` to use `suppress_if` in reference to an `assert_call` check, which I have been thinking about doing.

lib/elixir_analyzer/test_suite/newsletter.ex
jiegillet: Easy enough!

lib/elixir_analyzer/test_suite/newsletter.ex
angelikatyborska: >  It would also allow a feature to use suppress_if in reference to an assert_call check, which I have been thinking about doing.

That's an excellent idea, I noticed that that's missing too...

I created a ticket for this as a reminder 🙂

PR #133, opened 2021-09-03
[#130] Add a common check for IO.inspect

### LINKED ISSUE #130: Add a common check for `IO.inspect`
Some students don't pay too much attention to cleaning up their code before submitting it for mentoring. They leave around commented out code, nonsense comments, and debug `IO.inspect` calls.

It would be nice if we could nudge the students to clean up their solution with the analyzer. It should be possible to at least detect `IO.inspect` calls. Telling apart good comments from nonsense comments is probably not possible automatically 🙂.

Please note that there are exercises that require `IO.inspect` calls! Definitely the concept exercise `rpg-character-sheet` that teaches `IO` does. We would need a way to exclude those from the check.

jiegillet:
Closes #130.
This PR introduces two features

1. **A new common check for `IO.Inspect`**

This common check uses a `assert_call` under the hood to enjoy its module tracking features. 
This means that I had to add an option 
```elixir
use ElixirAnalyzer.ExerciseTest, suppress_common_tests: true
```
 because every test module runs `assert_call`, `feature` and the common checks. Since this is a common check, it would keep looping forever. 
 
Because some exercises require `IO.inspect` (`rpg-character-sheet`), I also introduced another option 
```elixir
use ElixirAnalyzer.ExerciseTest,  suppress_tests: [Constants.solution_debug_functions()]
``` 
which can suppress any type given its final comment. If that is too broad, it shouldn't be hard to suppress by the test name instead of comment (several tests can share one comment).

It would be very easy to add `IO.puts/1` to the list of debug functions. Let me know.

2. **assert_call/feature/common check cross suppress_if**

Now `feature` and `assert_call` can be suppressed by the other and by common checks. Since `suppress_if` works by checking test names, I had to add names for all common checks and test files. I picked names identical to the comments for simplicity.

### DIFF
--- a/lib/elixir_analyzer/exercise_test/common_checks/debug_functions.ex
+++ b/lib/elixir_analyzer/exercise_test/common_checks/debug_functions.ex
@@ -10,7 +10,7 @@ defmodule ElixirAnalyzer.ExerciseTest.CommonChecks.DebugFunctions do
   use ElixirAnalyzer.ExerciseTest, suppress_common_tests: true
 
   assert_no_call "solution doesn't use IO.inspect" do
-    type :actionable
+    type :informational
     comment ElixirAnalyzer.Constants.solution_debug_functions()
     called_fn module: IO, name: :inspect
   end
--- a/lib/elixir_analyzer/exercise_test/common_checks/debug_functions.ex
+++ b/lib/elixir_analyzer/exercise_test/common_checks/debug_functions.ex
@@ -1,0 +1,32 @@
+defmodule ElixirAnalyzer.ExerciseTest.CommonChecks.DebugFunctions do
+  @moduledoc """
+  This is an exercise analyzer extension module used for common tests lookin for debugging functions
+  """
+
+  alias ElixirAnalyzer.Comment
+  alias ElixirAnalyzer.Submission
+
+  # Suppressing common tests prevents this module to recursively call common tests
+  use ElixirAnalyzer.ExerciseTest, suppress_common_tests: true
+
+  assert_no_call "solution doesn't use IO.inspect" do
+    type :actionable
+    comment ElixirAnalyzer.Constants.solution_debug_functions()
+    called_fn module: IO, name: :inspect
+  end
+
+  @spec run(String.t()) :: [{:fail, %Comment{}}]
+  def run(code_as_string) do
+    %Submission{comments: comments} =
+      %Submission{code_file: "", code_path: "", path: "", analysis_module: __MODULE__}
+      |> __MODULE__.analyze(code_as_string)
+
+    case comments do
+      [%{comment: comment, type: type} | _] ->
+        [{:fail, %Comment{comment: comment, type: type, name: comment}}]
+
+      _ ->
+        []
+    end
+  end
+end
--- a/lib/elixir_analyzer/exercise_test/common_checks/debug_functions.ex
+++ b/lib/elixir_analyzer/exercise_test/common_checks/debug_functions.ex
@@ -3,30 +3,13 @@ defmodule ElixirAnalyzer.ExerciseTest.CommonChecks.DebugFunctions do
   This is an exercise analyzer extension module used for common tests lookin for debugging functions
   """
 
-  alias ElixirAnalyzer.Comment
-  alias ElixirAnalyzer.Submission
-
-  # Suppressing common tests prevents this module to recursively call common tests
-  use ElixirAnalyzer.ExerciseTest, suppress_common_tests: true
-
-  assert_no_call "solution doesn't use IO.inspect" do
-    type :informational
-    comment ElixirAnalyzer.Constants.solution_debug_functions()
-    called_fn module: IO, name: :inspect
-  end
-
-  @spec run(String.t()) :: [{:fail, %Comment{}}]
-  def run(code_as_string) do
-    %Submission{comments: comments} =
-      %Submission{code_file: "", code_path: "", path: "", analysis_module: __MODULE__}
-      |> __MODULE__.analyze(code_as_string)
-
-    case comments do
-      [%{comment: comment, type: type} | _] ->
-        [{:fail, %Comment{comment: comment, type: type, name: comment}}]
-
-      _ ->
-        []
+  defmacro __using__(_opts) do
+    quote do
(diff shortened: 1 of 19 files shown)


test/support/analyzer_verification/suppress_if.ex
```diff
@@ -0,0 +1,59 @@
...
+      foo()
+    end
+  end
+
+  assert_no_call "assert 1: no foo() unless common check was found" do
+    comment "assert 1: foo() was called"
+    # This should work:
+    # suppress_if Constants.solution_debug_functions(), :fail
+    # but in assert_call, it gets treated as an AST, so the final string is required 
+    suppress_if "elixir.solution.debug_functions", :fail
```
jiegillet, line 24: This is an annoying bug. I will open an issue for it.
I think it is related to that other bug that doesn't let you use
```elixir
alias alias ElixirAnalyzer.Constants
assert_no_call "x" do
  comment Constants.exercise()
...
```
but forces you to use the full path for the comment
```elixir
assert_no_call "x" do
  comment ElixirAnalyzer.Constants.exercise()
...
```

lib/elixir_analyzer/exercise_test/common_checks/debug_functions.ex
```diff
@@ -0,0 +1,32 @@
+defmodule ElixirAnalyzer.ExerciseTest.CommonChecks.DebugFunctions do
+  @moduledoc """
+  This is an exercise analyzer extension module used for common tests lookin for debugging functions
+  """
+
+  alias ElixirAnalyzer.Comment
+  alias ElixirAnalyzer.Submission
+
+  # Suppressing common tests prevents this module to recursively call common tests
+  use ElixirAnalyzer.ExerciseTest, suppress_common_tests: true
```
angelikatyborska, line 10: This doesn't look so good... circular dependencies between modules usually mean trouble. Is there a way to avoid this while still using `assert_no_call` for this common check?

I do have an idea and I'm proposing it without checking if it will work, so beware 😁 could we inject the calls to `assert_no_call` via `__using__` in `ElixirAnalyzer.ExerciseTest`? This module could define for example a macro like this instead of the `run` function that could be then used in `ElixirAnalyzer.ExerciseTest`:

```elixir
defmacro inject() do
  quote do
    assert_no_call "solution doesn't use IO.inspect" do
	  type :informational
	  comment ElixirAnalyzer.Constants.solution_debug_functions()
	  called_fn module: IO, name: :inspect
	end
  end
end
```

test/support/analyzer_verification/suppress_if.ex
angelikatyborska: I once tried to debug it but we're doing so much metaprogramming that I gave up. Anyway, it's good to open that issue.

test/support/analyzer_verification/suppress_if.ex
```diff
@@ -0,0 +1,59 @@
+defmodule ElixirAnalyzer.Support.AnalyzerVerification.SuppressIf do
+  @moduledoc """
+  This is an analyzer extension module to test the option :suppress_if between features/assert_call/common checks
+  """
```
angelikatyborska, line 4: Awesome! You're adding all of the analyzer features that I wanted to have but not really implement them myself 🥰

lib/elixir_analyzer/exercise_test/common_checks/function_names.ex
```diff
@@ -26,6 +26,7 @@ defmodule ElixirAnalyzer.ExerciseTest.CommonChecks.FunctionNames do
         {:fail,
          %Comment{
            type: :actionable,
+           name: Constants.solution_function_name_snake_case(),
```
angelikatyborska, line 29: > I picked names identical to the comments for simplicity.
Good choice 👍

angelikatyborska:
I fixed the credo issues to see what the CI does next: there's a dialyzer issue and the comment text is still missing. Did I miss a website-copy PR?

jiegillet:
Thanks for the review!
I did not submit a PR on website-copy yet, I wanted to know if we wanted to add more debug functions, like `IO.puts` or maybe `IO.write`, and I knew this would not be merged in one go :)

lib/elixir_analyzer/exercise_test/common_checks/debug_functions.ex
jiegillet: Mmh, I didn't think about circular dependencies... True, it's not desirable.

I had another version working which didn't do that, but it involved basically reimplementing `ElixirAnalyzer.ExerciseTest ` in the file. It worked but wasn't pretty. 

I think I get what you're proposing. Something like this?

```elixir
defmodule ElixirAnalyzer.ExerciseTest do
  alias  [...]

  defmacro __using__(opts) do
    quote do
      use ElixirAnalyzer.ExerciseTest.CommonChecks # <== this injects extra features and calls 
      use ElixirAnalyzer.ExerciseTest.Feature
      use ElixirAnalyzer.ExerciseTest.AssertCall
      @suppress_tests unquote(opts)[:suppress_tests]

[...]
```
I'm too sleepy to investigate more right now, but I'll try soon :)

(PS: doesn't the GitHub highlighter know that `...` is a valid variable name? haha)

test/support/analyzer_verification/suppress_if.ex
jiegillet: I had never really done any metaprogramming before contributing here... Talk about a baptism by fire!

angelikatyborska:
Ah, right. I think `IO.inspect` is enough. The other functions are not well suited for debugging and I don't see them used much for that.

lib/elixir_analyzer/exercise_test/common_checks/debug_functions.ex
angelikatyborska: Yeah, something like it. I guess we need some sort of split between common checks that work directly on the AST and common checks that inject `form` and `assert_call` macros.

jiegillet: I came up with something relatively painlessly. It's using a lot of `__using__` macros, but it might be the price we have to pay? Unless you can think of something cleaner...

angelikatyborska: It looks perfect to me now

angelikatyborska (approved):

PR #138, opened 2021-09-04
[#134] `high-score` analysis should allow reset_score with remove + add player

### LINKED ISSUE #134: `high-score` analysis should allow reset_score with remove + add player
Currently, this solution produces a comment about missing usage of `@initial_score` in `reset_score`. This comment should not be produced if `reset_score` uses `add_player` because then it uses `@initial_score` implicitly.

It might be possible to use the `suppress_if` option to make that happen but I don't know if we can write it in a way that wouldn't produce any extra undesired comments.

```elixir
defmodule HighScore do
  @initial_score 0

  def new(), do: %{}

  def add_player(scores, name, score \\ @initial_score) do
    Map.put(scores, name, score)
  end

  def remove_player(scores, name) do
    Map.delete(scores, name)
  end

  def reset_score(scores, name) do
    scores |> remove_player(name) |> add_player(name)
  end

  def update_score(scores, name, score) do
    Map.update(scores, name, score, &(&1 + score))
  end

  def get_players(scores) do
    Map.keys(scores)
  end
end
```

jiegillet:
Fixes #134. 
Easy-peasy. Make both conditions suppress if the other passes.
I think this doesn't generalize to more than 2 conditions, though. If there is a need for that, we may need to generalize `assert_call` to make the following valid:
```elixir
  assert_call "uses the module attribute in reset_score or indirectly via add_player" do
    type :essential
    comment ElixirAnalyzer.Constants.high_score_use_module_attribute()
    find: any # :one, :all, :none ?
    
    call do
      calling_fn module: HighScore, name: :reset_score
      called_fn name: :@
    end

    call do
      calling_fn module: HighScore, name: :reset_score
      called_fn name: :add_player
    end
  end
```

### DIFF
--- a/lib/elixir_analyzer/test_suite/high_score.ex
+++ b/lib/elixir_analyzer/test_suite/high_score.ex
@@ -32,5 +32,14 @@ defmodule ElixirAnalyzer.TestSuite.HighScore do
     comment ElixirAnalyzer.Constants.high_score_use_module_attribute()
     calling_fn module: HighScore, name: :reset_score
     called_fn name: :@
+    suppress_if "uses add_player in reset_score", :pass
+  end
+
+  assert_call "uses add_player in reset_score" do
+    type :essential
+    comment ElixirAnalyzer.Constants.high_score_use_module_attribute()
+    calling_fn module: HighScore, name: :reset_score
+    called_fn name: :add_player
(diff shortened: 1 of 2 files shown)


angelikatyborska (approved):
That's a relief. Perfect!

PR #140, opened 2021-09-04
[#139] Bug -- `RPG.CharacterSheet` false-positive error

### LINKED ISSUE #139: Bug -- `RPG.CharacterSheet` false-positive error
Essential-level error is raised despite meeting the requirements

Screenshot from 2021-09-04 00-26-49

jiegillet:
Fixes #139.

This was fairly tricky, please review with care.

Because the case raised by @neenjaw is likely to happen regularly, I chose to treat pipes as a special case throughout all `feature` matches, not just for `_block_ends_with`. I came up with those rules:

1. If there is a pipe in the `form`, there must be a pipe in the code.
2. If there is no pipe in the `form` but a pipe in the code, the pipe in the code will get "unpiped" go on with the match.

I dealt with one edge case: even `mix format` allows both `:ok |> foo` and `:ok |> foo()` so I made sure that didn't matter for the `form` or the code.

I noticed another edge case: if the form is `foo(_ignore)`, then `name |> foo(:ok)` will unfortunately match, because `feature` goes down the AST and tries to match beyond the pipe. So this PR is not rainbow-unicorn-magic  (🌈🦄🧙‍♀️) good, although to be fair, this is the current behavior anyway, so no existing test can break from this.

I added tests for all of these points.

### DIFF
--- a/test/elixir_analyzer/exercise_test/feature/pipes_test.exs
+++ b/test/elixir_analyzer/exercise_test/feature/pipes_test.exs
@@ -1,0 +1,184 @@
+defmodule ElixirAnalyzer.ExerciseTest.Feature.PipesTest do
+  use ElixirAnalyzer.ExerciseTestCase,
+    exercise_test_module: ElixirAnalyzer.Support.AnalyzerVerification.Feature.Pipes
+
+  test_exercise_analysis "function with one parameter",
+    comments_exclude: ["not one parameter"] do
+    [
+      defmodule MyModule do
+        def hi(name) do
+          foo(name)
+        end
+      end,
+      # This case matches, because feature looks down the AST
+      defmodule MyModule do
+        def hi(name) do
+          name |> foo(:ok)
+        end
+      end,
+      defmodule MyModule do
+        def hi(name) do
+          name |> foo
+        end
+      end,
+      defmodule MyModule do
+        def hi(name) do
+          name |> foo()
+        end
+      end,
+      defmodule MyModule do
+        def hi(name) do
+          name |> String.downcase() |> foo()
+        end
+      end,
+      defmodule MyModule do
+        def hi(name) do
+          name |> String.downcase() |> foo
+        end
+      end
+    ]
+  end
+
+  test_exercise_analysis "function not with one parameter",
+    comments_include: ["not one parameter"] do
+    [
+      defmodule MyModule do
+        def hi(name) do
+          foo()
+        end
+      end,
+      defmodule MyModule do
+        def hi(name) do
+          foo
+        end
+      end
+    ]
+  end
+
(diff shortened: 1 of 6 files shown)


test/elixir_analyzer/exercise_test/feature/pipes_test.exs
```diff
@@ -0,0 +1,184 @@
...
+        def hi(name) do
+          foo(name)
+        end
+      end,
+      # This case matches, because feature looks down the AST
+      defmodule MyModule do
+        def hi(name) do
+          name |> foo(:ok)
+        end
+      end,
```
angelikatyborska, line 18: Do we have any tests that depend on this behavior?

test/elixir_analyzer/exercise_test/feature/pipes_test.exs
```diff
@@ -0,0 +1,184 @@
...
+      defmodule MyModule do
+        def hi(name) do
+          name |> String.downcase() |> foo
+        end
+      end
+    ]
+  end
+
+  test_exercise_analysis "function not with one piped parameter",
+    comments_include: ["not one piped parameter", "not one piped parameter (no parens)"] do
```
angelikatyborska, line 133: This is question to make sure I understand the implementation: those two forms (with parens, without parens) will always be either both matched or both unmatched, right? Because we are adding the parens. So you cannot write a test that will pass this?
```elixir
  test_exercise_analysis "...",
    comments_exclude: ["not one piped parameter"],
    comments_include: ["not one piped parameter (no parens)"] do
```

angelikatyborska (approved):
I read everything and played around by modifying tests. Everything looks sane to me. Good job!

I left some questions to verify if I understood correctly, but I am also going to merge this now because it's a pretty bad bug.

test/elixir_analyzer/exercise_test/feature/pipes_test.exs
jiegillet: I don't think so. Not in the ones I made anyway. 

This would be an issue if you were trying to match a function of arity 1, you could get a false positive that matches the same function of arity 2 used with a pipe. I have no idea how likely that is to happen organically. To avoid that you would have say match `foo(_ignore)` but not `_ignore  |> foo(_ignore)`.

I cannot think of a way to avoid this at the moment. At least we are aware of it.

test/elixir_analyzer/exercise_test/feature/pipes_test.exs
jiegillet: Exactly. 
They do exactly the same (but do not trigger the compiler error saying they are the same form, because the AST are a bit different).

PR #145, opened 2021-09-11
[#137] `assert_call` detects indirect calls from helper functions

### LINKED ISSUE #137: `assert_call` doesn't see "through" nested function calls
image

I added a helper function to re-use code, and got a recommendation to re-use code in a different way.

I understand this is the current limitation of the macro-based approach. Not sure if it could be improved in general, or maybe just for this exercise.

Alternative approach would be to use tracing. Here I adapted a `FreelancerRates` test to check that function is re-used, and this approach works even if additional helper functions are introduced:

```elixir
    @tag task_id: 4
    test "calls apply_discount" do
      :erlang.trace(self(), true, [:call])
      :erlang.trace_pattern({FreelancerRates, :apply_discount, 2}, true, [:call_count])

      assert FreelancerRates.days_in_budget(1_600, 50, 0.0) == 4

      {:call_count, apply_discount_calls} = :erlang.trace_info({FreelancerRates, :apply_discount, 2}, :call_count)
      assert apply_discount_calls == 1

      :erlang.trace_pattern({FreelancerRates, :apply_discount, 2}, false, [:call_count])
      :erlang.trace(self(), false, [:call])
    end
```

jiegillet:
This PR introduces a new feature to `assert_call` and closes #137.

Now, if a `calling_fn` is specified, `assert_call ` will track all the functions called in that function and check if any of those call `called_fn`.

This is helpful if people use helper functions. While working on this, I kept wondering if it might not have undesirable effects, but none of the existing unit tests broke (except for 3 specific `assert_call` tests that specified the behavior) so that put me at ease, and now I feel that it's worth having.

It would be simple to add flags that turn off the behavior or limit the nesting depth, but that might make things more complicated for people writing checks. 

I'm happy to discuss further.

### DIFF
--- a/lib/elixir_analyzer/exercise_test/assert_call/compiler.ex
+++ b/lib/elixir_analyzer/exercise_test/assert_call/compiler.ex
@@ -71,7 +71,7 @@ defmodule ElixirAnalyzer.ExerciseTest.AssertCall.Compiler do
   """
   @spec handle_traverse_result({any, map()}) :: boolean
   def handle_traverse_result({_, %{found_called: found, calling_fn: calling_fn} = acc}) do
-    found || (!is_nil(calling_fn) && indirect_call?(acc))
+    found or (not is_nil(calling_fn) and indirect_call?(acc))
   end
 
   @doc """
@@ -404,24 +404,23 @@ defmodule ElixirAnalyzer.ExerciseTest.AssertCall.Compiler do
 
   # Check if a function was called through helper functions
   def indirect_call?(%{called_fn: called_fn, calling_fn: calling_fn, function_call_tree: tree}) do
-    #    IO.inspect({called_fn, calling_fn, tree})
-
     cond do
+      # calling_fn wasn't defined in the code, or was searched already
       is_nil(tree[calling_fn]) ->
         false
 
+      # calling_fn directly called called_fn
       called_fn in tree[calling_fn] ->
         true
 
+      # calling_fn didn't call called_fn, recursively check if other called functions did
       true ->
-        tree[calling_fn]
-        # Delete function itself once because it is always added from the `def` node
-        |> List.delete(calling_fn)
-        |> Enum.any?(
+        Enum.any?(
+          tree[calling_fn],
           &indirect_call?(%{
             called_fn: called_fn,
             calling_fn: &1,
-            # Remove tree branch already explored in case of cyclical calls
+            # Remove tree branch since we know it doesn't call called_fn
             function_call_tree: Map.delete(tree, calling_fn)
           })
         )
--- a/lib/elixir_analyzer/exercise_test/assert_call/compiler.ex
+++ b/lib/elixir_analyzer/exercise_test/assert_call/compiler.ex
@@ -57,7 +57,8 @@ defmodule ElixirAnalyzer.ExerciseTest.AssertCall.Compiler do
       modules_in_scope: %{},
       found_called: false,
       called_fn: called_fn,
-      calling_fn: calling_fn
+      calling_fn: calling_fn,
+      function_call_tree: %{}
     }
 
     ast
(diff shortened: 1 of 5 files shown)


lib/elixir_analyzer/exercise_test/assert_call/compiler.ex
```diff
@@ -373,4 +383,48 @@ defmodule ElixirAnalyzer.ExerciseTest.AssertCall.Compiler do
...
+        |> List.delete(calling_fn)
+        |> Enum.any?(
+          &indirect_call?(%{
+            called_fn: called_fn,
+            calling_fn: &1,
+            # Remove tree branch already explored in case of cyclical calls
+            function_call_tree: Map.delete(tree, calling_fn)
+          })
+        )
+    end
```
angelikatyborska, line 428: It took me a while to understand those branches. Maybe some comments would help? Also, do those comments make sense, did I understand correctly?
```suggestion
    cond do
      # calling_fn was never even called
      is_nil(tree[calling_fn]) ->
        false
	  
      # called_fn was called directly by calling_fn
      called_fn in tree[calling_fn] ->
        true

      # called_fn wasn't called directly by calling_fn, look for indirect calls
      true ->
        tree[calling_fn]
        # Delete function itself once because it is always added from the `def` node
        |> List.delete(calling_fn)
        |> Enum.any?(
          &indirect_call?(%{
            called_fn: called_fn,
            calling_fn: &1,
            # Remove tree branch already explored in case of cyclical calls
            function_call_tree: Map.delete(tree, calling_fn)
          })
        )
    end
```

lib/elixir_analyzer/exercise_test/assert_call/compiler.ex
```diff
@@ -69,7 +70,9 @@ defmodule ElixirAnalyzer.ExerciseTest.AssertCall.Compiler do
   Handle the final result from the assert function
   """
   @spec handle_traverse_result({any, map()}) :: boolean
-  def handle_traverse_result({_, %{found_called: found}}), do: found
+  def handle_traverse_result({_, %{found_called: found, calling_fn: calling_fn} = acc}) do
+    found || (!is_nil(calling_fn) && indirect_call?(acc))
```
angelikatyborska, line 74: What's the difference between the situation when `found` is truthy vs when `called_fn in tree[calling_fn]` is truthy in `indirect_call?`?

angelikatyborska (approved):
Great job! This will be an enormous improvement to the accuracy of those checks. I think the behavior to follow indirect calls is always desired.

Do we need additional tests for `assert_no_call`? I looked through the existing usage and none of the exercise-specific checks specify `calling_fn` with using `assert_no_call`, so that case is not being tested anywhere.

lib/elixir_analyzer/exercise_test/assert_call/compiler.ex
jiegillet: You got the idea, but because `indirect_call?` calls itself recursively, it's a little bit more subtle I guess? I will add comments.

jiegillet (commented):
> Do we need additional tests for assert_no_call? I looked through the existing usage and none of the exercise-specific checks specify calling_fn with using assert_no_call, so that case is not being tested anywhere.

I don't think it's necessary, but I still lack experience with unit tests, so let me explain my train of thought.
The property "`assert_call` gives the inverse result to `assert_no_call`" is independent from the change in this PR, so it shouldn't really be affected and should be tested in a different place. If I had to re-check every possible interaction of properties for each new feature, the unit tests would grow exponentially. Of course, this is based on my belief that I do not modify existing properties when I add a new feature, which could be wrong obviously.
What is considered good practice for this? Keeping features separate? Mixing all features in complex unit tests systematically or only once in a while? 
Do you have experience with property testing? It's more complex but really cool. A quick search shows a few available libraries.

angelikatyborska:
Alrighty. Fine with me, good to merge from my side. Exciting!

> Of course, this is based on my belief that I do not modify existing properties when I add a new feature, which could be wrong obviously.
What is considered good practice for this? Keeping features separate? Mixing all features in complex unit tests systematically or only once in a while?

🙈 I have no idea. This is one of the most complex projects I have worked on testing-wise. Usually I have a 1 to 1 mapping of module - test module, but here it's not possible, it's so complex. I think I need to put https://pragprog.com/titles/lmelixir/testing-elixir/ on my reading list.

> Do you have experience with property testing?
No, zero.

PR #147, opened 2021-09-12
[#146] Missing parenthesis in zero-arity function definitions cause false-positives

### LINKED ISSUE #146: Missing parenthesis in zero-arity function definitions cause false-positves
This solution, while being totally correct, trigger a bunch of checks. The cause is no parenthesis in zero-arity function definitions. Skipping them is not an error and not even a warning, so we should fix this in the analyzer.
```elixir
  @moduledoc """
  A collection of functions related to creating a character sheet. 
  """

  @doc """
  Prints out the welcome message.
  """
  @spec welcome() :: binary()

  def welcome do
    IO.puts "Welcome! Let's fill out your character sheet together."
  end

  @doc """
  Gets the character name via input.
  """
  @spec ask_name() :: binary()

  def ask_name do
    ask("What is your character's name?\n")
  end

  @doc """
  Gets the character class via input.
  """
  @spec ask_class() :: binary()

  def ask_class do
    ask("What is your character's class?\n")
  end

  @doc """
  Gets the character's level via input, returns an integer. 
  """
  @spec ask_level() :: pos_integer()

  def ask_level do
    "What is your character's level?\n"
    |> ask()
    |> String.to_integer(10)
  end

  @doc """
  Runs the character sheet creation workflow.
  """
  @spec run() :: %{name: binary(), class: binary(), level: pos_integer()}

  def run do
    welcome()

    %{
      name: ask_name(), 
      class: ask_class(), 
      level: ask_level()
    }
    |> IO.inspect(label: "Your character")
  end

  # Given a question, asks it and trims leading & trailing whitespaces.
  defp ask(question) do
    question
    |> IO.gets()
    |> String.trim()
  end
end
```

Screenshot 2021-09-11 at 15-41-28 Exercism

jiegillet:
Fixes #146.

From this PR, `a` will always be treated to be equivalent with `a()` in a `feature`, and match both `a` and `a()` in code.

There was already some code that did this for the special case of pipes, but I removed it since this is more general.

### DIFF
--- a/lib/elixir_analyzer/exercise_test/feature/compiler.ex
+++ b/lib/elixir_analyzer/exercise_test/feature/compiler.ex
@@ -225,25 +225,24 @@ defmodule ElixirAnalyzer.ExerciseTest.Feature.Compiler do
     end
   end
 
+  # functions called with or without parentheses should match
+  def form_match?({function, form_params, []}, {function, params, atom}) when is_atom(atom) do
+    form_match?(form_params, params)
+  end
+
+  def form_match?({function, form_params, atom}, {function, params, []}) when is_atom(atom) do
+    form_match?(form_params, params)
+  end
+
   # Pipes are a special case, when pipes are in the form, they must be in the code
-  def form_match?({:|>, form_meta, [form_params, form_function]}, line) do
-    case line do
-      {:|>, meta, [params, function]} ->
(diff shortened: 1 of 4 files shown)


angelikatyborska (approved):
Perfect!

PR #154, opened 2021-09-14
Add `Map.update/4` check to `high-score`

### LINKED ISSUE #135: `high-score` analysis should require Map.update in `update_score`
A common mistake is to write `Map.get` + `Map.put` instead of `Map.update`. Add an analyzer check that assures that `Map.update` will be called in `update_score`.

`Map.update!` is not an option, it's excluded by the unit tests already.

Br1ght0ne:
Close #135. Sibling PR is exercism/website-copy#2072.


angelikatyborska (approved):
Looks great!

I'll wait for the website-copy PR first and rerun the CI before merging to be 100% sure that the comment works.

Issue #132, opened 2021-09-03
Celebratory comment when concept exercise solution 100% matches the exemplar


angelikatyborska:
Could we somehow add an analysis step to every concept exercise that will give a celebratory comment when your solution's AST matches 100% the exemplar solution?

It would be nice if we didn't have to copy-paste all of the exemplars to this repo, but somehow fetch them from the track repo. By git submodules?

Would that require adding some CI magic to periodically update the state of the submodule when the track changes so that the analyzer can be deployed often with the new exemplars?

jiegillet:
For most of the concept exercises, we already have the exemplar solution code in the tests, maybe it's simpler that way. Although I do agree, it would be super cool to fetch them automatically. My CI skills are close to 0 unfortunately, but I'm always looking out for things to learn ;)

jiegillet:
Exciting, let's do it :)

jiegillet:
I got the submodule and and automated syncing GitHub Action to work on my forks of `elixir-analyzer` and `elixir` 🎉 
I followed the workflows accepted in the SO link you provided, so that `elixir-analyzer` automatically syncs the submodule whenever a commit is pushed to the `main` branch of `elixir`. 

I still have some uncertainty around creating the GitHub Action tokens for `exercism` rather than for my own account, but here is what I did:

1. Create submodule in `elixir-analyzer` and commit [this file](https://github.com/jiegillet/elixir-analyzer/blob/jie-submodule/.gitmodules)
2. Create Github Action workflow in `elixir-analyzer` and commit [this file](https://github.com/jiegillet/elixir-analyzer/blob/jie-submodule/.github/workflows/elixir_submodule_sync.yml)
3. [Create a Private Access Token](https://docs.github.com/en/github/authenticating-to-github/keeping-your-account-and-data-secure/creating-a-personal-access-token) for my account, select permission "workflow". (this step probably won't work for exercism repo)
4. Copy token, it will need to be pasted into 2 repos
5. `elixir-analyzer` repo > Settings > Secrets > New repository secret > Name: SUBMODULE_TOKEN value: paste token from step 4. (I don't have access to settings for exercism repo)
6. `elixir-analyzer` repo > Actions > Elixir Submodule Sync > run workflow manually
7. If step 6 works and submodule points to latest commit, rejoice. Otherwise GOTO step `:rand.uniform(6)`.
[...]

PR #157, opened 2021-09-16
Add common check: solution identical to exemplar 

jiegillet:
Covers some features discussed in #132.

This PR:
- Introduces a git submodule of the `elixir` repo
- Reads the exemplar solutions to concepts exercises in `elixir` and compares it to the code, if they are identical, a celebratory comment is added

It does not
- Sync the submodule via GH Actions or otherwise
- Check that the exemplar solution passes all the tests

This is quite a lot of code, and I have some questions, so this is probably not ready to merge as is. I will leave my questions as comments so it's easier to discuss.


test/support/exercise_test_case.ex
```diff
@@ -147,4 +151,21 @@ defmodule ElixirAnalyzer.ExerciseTestCase do
...
+           Enum.find(@exercise_config, fn
+             {_slug, %{analyzer_module: ^exercise_test_module}} -> true
+             _ -> false
+           end),
+         {:ok, config_file} <-
+           [@concept_exercice_path, slug, @meta_config] |> Path.join() |> File.read(),
+         %{"files" => %{"exemplar" => [path]}} <- Jason.decode!(config_file),
+         exemplar_code <-
+           [@concept_exercice_path, slug, path] |> Path.join() |> File.read!() do
+      exemplar_code
```
jiegillet, line 166: Is this readable? I really squeezed the power of `with` here...

test/support/exercise_test_case.ex
```diff
@@ -147,4 +151,21 @@ defmodule ElixirAnalyzer.ExerciseTestCase do
   def assert_comments(_, _, _) do
     :noop
   end
+
+  def find_exemplar_code(exercise_test_module) do
+    with {slug, _module} <-
+           Enum.find(@exercise_config, fn
+             {_slug, %{analyzer_module: ^exercise_test_module}} -> true
+             _ -> false
```
jiegillet, line 159: This is not super efficient because I'm going through the whole list of modules to find the original slug...

jiegillet:
The test seem to fail, I guess it's because the submodule isn't loaded... Hopefully it will work locally, you need to run `git submodule update --remote --recursive`.

jiegillet:
I modified reading the exemplar file according to the discussion with Jeremy, then added tests. There are so many files modified now. Would you prefer I break it down in smaller PRs? 

I also realized that I think showing the students errors when the exemplar files are not found or parsed wrong is probably a bad idea, it's better to have it in the logs. We have access to these right?

lib/elixir_analyzer/exercise_test.ex
```diff
@@ -114,17 +121,31 @@ defmodule ElixirAnalyzer.ExerciseTest do
...
         |> Submission.halt()
-        |> Submission.append_comment(%Comment{
-          comment: Constants.general_parsing_error(),
-          params: comment_params,
-          type: :essential
-        })
+        |> Submission.append_comment(
+          if exemplar do
+            %Comment{
+              comment: Constants.general_exemplar_parsing_error(),
```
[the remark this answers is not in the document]
jiegillet: Update: I now thing writing it in the log is better.

angelikatyborska:
I'm sorry, I still don't have the time and energy to do a full review, but I'll try to answer some of the questions now.

> I also realized that I think showing the students errors when the exemplar files are not found or parsed wrong is probably a bad idea

I agree.

>  it's better to have it in the logs. We have access to these right?

In theory yes. In practice we only have access to an ad hoc page for maintainers where the logs are visible for a few minutes and then they get pushed out by metadata from other submissions. Only a limited number of most recent submissions are shown, and that limit is platform-wide.

So in practice you need to run a problematic submission and take a look at that page within a minute, otherwise you won't find the log again. I really hope that's something that will be resolved soon. Better debugging tools for maintainers would allow a lot of growth and improvements in all the tracks (cc @iHiD).

> I modified reading the exemplar file according to the discussion with Jeremy

I think this also means we need to get rid of the git submodule. Is that possible? Jeremy said he doesn't want analyzer deployments if nothing functional changes in the analyzer. And every commit to main does trigger a deployment, including updating the reference to the submodule, right?

> Would you prefer I break it down in smaller PRs?

No need, if they're all part of this one big feature, it's fine.

lib/elixir_analyzer/exercise_test/common_checks/exemplar_comparaison.ex
```diff
@@ -0,0 +1,27 @@
...
+  """
+
+  alias ElixirAnalyzer.Constants
+  alias ElixirAnalyzer.Comment
+
+  @spec run(Macro.t(), nil | Macro.t()) :: [{:pass | :fail | :skip, %Comment{}}]
+  def run(_ast, nil), do: []
+
+  def run(code_ast, exemplar_ast) do
+    if Macro.to_string(code_ast) == Macro.to_string(exemplar_ast) do
```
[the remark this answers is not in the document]
angelikatyborska: Both get turned into ASTs first, so differences like single-line `if` vs multiline `if` won't matter, right? That's still less strict than doing a strict string equality on the original sources :)

Btw. is it necessary to turn them back into strings for the comparison?

iHiD:
> So in practice you need to run a problematic submission and take a look at that page within a minute, otherwise you won't find the log again. I really hope that's something that will be resolved soon. Better debugging tools for maintainers would allow a lot of growth and improvements in all the tracks (cc @iHiD).

Agreed. There's a very long TODO list atm. Need to find some money, hire some people, and move faster :)

lib/elixir_analyzer/exercise_test/common_checks/exemplar_comparaison.ex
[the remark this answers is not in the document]
jiegillet: Yes, I agree it is not the absolute strictest check, but it's still pretty strict, all variable names need to be the same, same typespecs and docs (those are not too much of an issue for concept exercise, I know), etc.

It's not strictly speaking necessary to compare strings, at first I was going through both ASTs and neglecting the metadata, but I kept forgetting corner cases... For debugging I was printing the code as strings and I realized I might as well do that. It's one function rather than 25+ lines of code.

jiegillet:
I'll modify to use logs then.

> I think this also means we need to get rid of the git submodule. Is that possible? Jeremy said he doesn't want analyzer deployments if nothing functional changes in the analyzer. And every commit to main does trigger a deployment, including updating the reference to the submodule, right?

I'm still using the submodule for testing all the analyzer checks, since short of copy-pasting, that's the only way we can compare the the exemplar solution. In that sense, we could use the CI to update the submodule on `push` on `elixir-analyzer main`, this way it would only update whenever the analyzer would be deploying anyway.

test/support/exercise_test_case.ex
angelikatyborska: I guess this for our own unit testing purposes that doesn't matter much.

config/test.exs
```diff
@@ -1,3 +1,3 @@
 use Mix.Config
 
-config :logger, level: :error
+config :logger, level: :warn
```
jiegillet, line 3: Without this I could not capture the logs I wanted to test.

test/support/exercise_test_case.ex
angelikatyborska: It is not 😁 I think the biggest problem are the multi-step clauses that don't fit on a single line. The mixture of `File.read` and `File.read!` is also confusing to me.  If this is for testing purposes only, can't everything just raise an error?

lib/elixir_analyzer.ex
```diff
@@ -43,8 +43,8 @@ defmodule ElixirAnalyzer do
   * `:output_file`, - specifies the name of the output_file, defaults to
     `@output_file` (`analysis.json`)
 
-  * `:exercise_config` - specifies the path to the JSON exercise configuration,
-    defaults to `@exercise_config` (`./config/exercise_data.json`)
+  * `:exercise_config` - specifies the path to the exercise configuration,
+    defaults to `@exercise_config` (`./config/config.exs`)
```
angelikatyborska, line 47: 👍  old info

angelikatyborska:
>  we could use the CI to update the submodule on push on elixir-analyzer main, this way it would only update whenever the analyzer would be deploying anyway.

Can this be achieved _before_ making the commit on `main`? Because if we do it _after_ the commit is pushed to main, there will be two deployments :(

test_data/lasagna/missing_exemplar/.meta/design.md
```diff
@@ -0,0 +1,47 @@
+# Design
```
angelikatyborska, line 1: The design files in the test data lasagnas are probably not needed

lib/elixir_analyzer.ex
```diff
@@ -166,7 +168,14 @@ defmodule ElixirAnalyzer do
     code_path = Path.dirname(full_code_path)
     code_file = Path.basename(full_code_path)
 
-    {code_path, code_file, exercise_config[:analyzer_module] || ElixirAnalyzer.TestSuite.Default}
+    exemplar_path =
+      case meta_config["files"]["exemplar"] do
+        [path | _] -> Path.join(params.path, path)
+        _ -> nil
```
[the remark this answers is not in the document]
jiegillet: A few line above there was already something similar
```elixir
relative_code_path = meta_config["files"]["solution"] |> hd()
```
This ignores anything but the first file and crash if it's empty.

test/support/exercise_test_case.ex
jiegillet: Excellent point.

jiegillet: Actually, I spoke too fast, this mustn't crash because it has to return `nil` for practice exercises.
I'll still rewrite it.

jiegillet: I'm struggling with this. Every `!` has a precise reason for being there or not. I settled for adding comments for the moment.

jiegillet:
> Can this be achieved _before_ making the commit on `main`? Because if we do it _after_ the commit is pushed to main, there will be two deployments :(

Really? It would make sense to me that the CI would be run on the push, but before the deployment. For example you may want to have an action that prevents the deployment in some cases. This is just guesswork though, I'll look into it.

iHiD:
Deployment is handled through the GHA on push, so what angelika says is correct. However, presuming the submodule is being ignored in the dockerfile (haven't checked, but it should be) the second run should be a noop. But it would need enough time between the two for that to happen.

Another option would be to update submodules automatically as a PR CI, so it happens on the pull-request action instead where it adds a new commit. It's a messy option though.

jiegillet:
I see.
Would adding a step in the deployment action be an option? I'm assuming you want to keep those identical from repo to repo though.

Another option would be to have this is in the test CI. Maybe we don't need to commit and push the submodule sync, just checking it out would be enough. If we want to run tests locally, we update the submodule locally and commit that along PRs... A bit opaque maybe.

Yet another option is to do without submodules and checking the exemplar from the repo URL, we do that for checking comments exist.

angelikatyborska:
That's a hard dilemma. I'm not sure what's the best course of action here.

This sounds reasonable:

> Another option would be to have this is in the test CI. Maybe we don't need to commit and push the submodule sync, just checking it out would be enough. If we want to run tests locally, we update the submodule locally and commit that along PRs... A bit opaque maybe.

But it might cause a situation where the analyzer returns both a celebratory comment and a negative comment. This would happen if we update the analyzer to do some new check but forgot to verify the new check against the most up to date exemplar. Sounds unlikely though, so maybe this is the best option?

This sounds the cleanest and safest:

> Yet another option is to do without submodules and checking the exemplar from the repo URL, we do that for checking comments exist.

Except that it will significantly slow down the tests a lot if each test needs to pull the exemplar, or slow down the tests slightly if we git clone the repo once when starting the tests.

jiegillet:
> But it might cause a situation where the analyzer returns both a celebratory comment and a negative comment. This would happen if we update the analyzer to do some new check but forgot to verify the new check against the most up to date exemplar. Sounds unlikely though, so maybe this is the best option?

This could happen locally, but then the CI would catch it. That's what it's for. I forget to run `mix credo` almost all of the time but the CI never forgets :)

> Except that it will significantly slow down the tests a lot if each test needs to pull the exemplar, or slow down the tests slightly if we git clone the repo once when starting the tests.

Not a huge deal IMO, the tests on `elixir` take like 5 minutes. We tag them with `:external` so they are not a bother locally. 

I know this doesn't help much but I don't really have a preference among all these options...

iHiD:
I'm unsubscribing from this. Re-tag me if you want my thoughts on anything 🙂

angelikatyborska:
> This could happen locally, but then the CI would catch it.

Then let's go for that option. To sum up, this is what I think that means:

- There is a git submodule for the track repo here.
- The submodule is used in unit tests for the new "identical to exemplar" check.
- The submodule is not used in production checks because the exemplar is passed as part of the solution.
- The submodule doesn't get updated automatically in any way to prevent unnecessary deployments.
- On CI, the submodule gets updated before running unit tests to ensure we test against the newest exercise versions, but then the changes are discarded.
- If the CI run failed for the above reason, the dev making the PR is expected to also update the submodule reference in their PR.

jiegillet:
That's exactly what I think that means too. 
OK then, I'll get started on the CI checks! I think the rest of the code is more or less OK. 

Messing with CI with my amount of experience will probably mean some amount of trial and error, so I apologize in advance if I end up spamming commits.

angelikatyborska:
Spamming commits is unfortunately The Only Way ™️ to work with GitHub workflows. I even have a dedicated repo for this for myself :D (https://github.com/angelikatyborska/gha-test)

jiegillet:
Looks like it's working. It only took 14 commits on my fork before I found the solution, which was literally adding `with:  submodules: true`. 🙈

angelikatyborska:
> which was literally adding with: submodules: true. 🙈

Is this also going to update the submodule to the latest commit?

jiegillet:
Te

jiegillet:
Sorry for previous message :)

> Is this also going to update the submodule to the latest commit?

Yes, I believe it does. I checked the log of the action and it runs
```
git submodule sync
git -c http.https://github.com.extraheader="AUTHORIZATION: basic ***" submodule update --init --force
```

lib/elixir_analyzer.ex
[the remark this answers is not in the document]
jiegillet: I thought about this a bit more.
For the exemplar code, I guess it's mostly fine because AFAIK, there is only one file for every exercise fo far, and unexpected situations (no files, multiple files) are very unlikely. 
For student code, that is more of an issue. Not for website submissions, since I don't think they can create new files, but for CLI submissions, there could be many (Tim loves 3+ files solutions). So the analysis would miss anything outside of the main file. 

Am I correct in saying that the directory/file structure in Elixir doesn't matter as long as it's in `lib`, and that only module names are important? 
If that is the case, a simple solution is to concatenate all the files that are in `lib` (as opposed to `meta_config["files"]["solution"]` which only points to the stub files) and pretend they were all one file. Actually we can do the same for the exemplar.

angelikatyborska:
I  tried to resolve the conficts and merge but I don't think that's straight-forward, could you do it?

I also noticed a small typo `ExemplarComparaison` -> `ExemplarComparison`

jiegillet:
On it!

angelikatyborska:
Let's see what happens when I press the merge button 😁

angelikatyborska (approved):

PR #159, opened 2021-09-21
Move raise_fn_clause_error check to common checks

### LINKED ISSUE #156: Common check for `raise FunctionClauseError`
Some of the unit tests for learning exercises (and collatz-conjecture) contain assertion that specific function calls raise `FunctionClauseError`. This is to check that the student didn't implement too permissive function clauses.

However, some students take that too literally and actually write `def foo(_), do: raise FunctionClauseError`.

There should be a common check for every solution that `raise FunctionClauseError` is never called. AFAIK there isn't ever a need to raise this error explicitly (at least on Exercism). 

It should be possible to implement this by using the `feature / form` macros, similar to how `ElixirAnalyzer.ExerciseTest.CommonChecks.DebugFunctions` was implemented with the `assert_no_call` macro.

Br1ght0ne:
Close #156.


lib/elixir_analyzer/exercise_test/common_checks/uncommon_errors.ex
```diff
@@ -0,0 +1,19 @@
...
+  defmacro __using__(_opts) do
+    quote do
+      feature "raises function clause error" do
+        find :none
+        type :actionable
+        comment ElixirAnalyzer.Constants.solution_raise_fn_clause_error()
+
+        form do
+          raise FunctionClauseError
+        end
```
Br1ght0ne, line 15: Any other forms that need to be handled?

angelikatyborska: One more. `raise` can also be used with structs:
```elixir
iex(6)> raise %FunctionClauseError{}
** (FunctionClauseError) no function clause matches
```

test/elixir_analyzer/test_suite/two_fer_test.exs
```diff
@@ -209,19 +209,4 @@ defmodule ElixirAnalyzer.TestSuite.TwoFerTest do
...
-        def two_fer(name \\ "you") do
-          if is_bitstring(name) do
-            "One for #{name}, one for me."
-          else
-            raise FunctionClauseError
-          end
-        end
-      end
-    end
-  end
```
angelikatyborska, line 226: It's a good choice to remove the exercise-specific test, but we need to have some sort of unit test for this check. Could you add something like in this PR https://github.com/exercism/elixir-analyzer/pull/153/files (`test/elixir_analyzer/exercise_test/common_checks/last_line_assignment_test.exs`)?

angelikatyborska (commented):
Nice! I wasn't aware that this check already exists for Two-Fer. The existing comment for this check is general so it will still work after moving the check to common checks 🙂

PR #162, opened 2021-09-22
Add basic Strain analysis

### LINKED ISSUE #41: Write a basic `strain` analysis
This exercise is supposed to be solved without using the `Enum` module. See:

- https://github.com/exercism/elixir/blob/main/exercises/practice/strain/README.md#L32-L34
- https://github.com/exercism/elixir/blob/main/exercises/practice/strain/.meta/example.ex#L6
- https://github.com/exercism/elixir/blob/main/exercises/practice/strain/.meta/example.ex#L26

I think an open question is whether list comprehensions should be considered not allowed?

Br1ght0ne:
Close #41.


lib/elixir_analyzer/test_suite/strain.ex
```diff
@@ -0,0 +1,37 @@
...
+    type :essential
+    called_fn module: Enum, name: :filter
+    comment ElixirAnalyzer.Constants.strain_use_recursion()
+  end
+
+  assert_no_call "does not call Enum.reject/2" do
+    type :essential
+    called_fn module: Enum, name: :reject
+    comment ElixirAnalyzer.Constants.strain_use_recursion()
+  end
```
angelikatyborska, line 18: Those functions could be recreated with `Enum.reduce` :(. If `strain` is really a recursion-only exercise, then we need to forbid the whole `Enum`, `List`, `Stream` modules and list comprehensions. Like for `bird-count`.

lib/elixir_analyzer/test_suite/strain.ex
```diff
@@ -0,0 +1,37 @@
...
+  @moduledoc """
+  This is an exercise analyzer extension module for the concept exercise Strain
+  """
+
+  use ElixirAnalyzer.ExerciseTest
+
+  assert_no_call "does not call Enum.filter/2" do
+    type :essential
+    called_fn module: Enum, name: :filter
+    comment ElixirAnalyzer.Constants.strain_use_recursion()
```
jiegillet, line 11: If this gets merged after #163, the comments can be called with an alias `Constants.strain_...`.

lib/elixir_analyzer/test_suite/strain.ex
angelikatyborska: Allowing only a single `Enum` function might be confusing 🤔 I guess the question is what we want from this exercise. If it's supposed to be an exercise of recursion, I guess `Enum.reverse` should not be allowed either and the example should be updated. @jiegillet @neenjaw opinions?

The most starred solution uses simple recursion (no accumulator, no reverse): https://exercism.org/tracks/elixir/exercises/strain/solutions/morgangrubb

jiegillet: Whenever the exercise is "recreate an `Enum` function", solutions I like are always either `reduce` + `reverse` (aka `foldr`) or plain recursion.
My own solution is very similar to the one you link, plain recursion. However I think I would not keep people away from using other `Enum` functions. So I vote to only reject `filter` and `reject`.

angelikatyborska: Ok then, let's keep this check like it is now. We just need the comment for the PR to be ready :)

PR #163, opened 2021-09-25
Fix `assert_call` AST arguments bug

### LINKED ISSUE #142: `assert_call` macro treats `comment` and `suppress_if` arguments as AST?
See:

https://github.com/exercism/elixir-analyzer/blob/8e879533ee112493f2eb19e5b30437fe879d3be0/test/support/analyzer_verification/suppress_if.ex#L19-L26

Also this doesn't work:

```elixir

  assert_call "use module attribute" do
    type :essential
    comment Constants.high_score_use_module_attribute()
    called_fn name: :@
  end
```
Instead, one needs to use the full module path `ElixirAnalyzer.Constants.high_score_use_module_attribute()`.

Both are most possibly related.

jiegillet:
Fixes #142.

This one took me a while, when I finally got it, I couldn't help myself and I aliased `Constants` everywhere.

The issue was that we were escaping the map containing all the `assert_call` data, so everything inside was escaped. Instead, I mimicked the approach from `feature`, changed the map into a keyword list and unquoted that. The keyword list is a proper AST it seems so it worked fine (the map was not a proper AST, so I couldn't unquote it directly). 

It might be easier to review both commits separately.

### DIFF
--- a/lib/elixir_analyzer/exercise_test/common_checks/uncommon_errors.ex
+++ b/lib/elixir_analyzer/exercise_test/common_checks/uncommon_errors.ex
@@ -2,13 +2,14 @@ defmodule ElixirAnalyzer.ExerciseTest.CommonChecks.UncommonErrors do
   @moduledoc """
   This is an exercise analyzer extension module used for common tests looking for uncommon errors raised
   """
+  alias ElixirAnalyzer.Constants
 
   defmacro __using__(_opts) do
     quote do
-      feature "raises function clause error" do
+      feature Constants.solution_raise_fn_clause_error() do
         find :none
         type :actionable
-        comment ElixirAnalyzer.Constants.solution_raise_fn_clause_error()
+        comment Constants.solution_raise_fn_clause_error()
 
         form do
           raise FunctionClauseError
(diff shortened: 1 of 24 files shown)


lib/elixir_analyzer/exercise_test/common_checks/uncommon_errors.ex
```diff
@@ -2,13 +2,14 @@ defmodule ElixirAnalyzer.ExerciseTest.CommonChecks.UncommonErrors do
   @moduledoc """
   This is an exercise analyzer extension module used for common tests looking for uncommon errors raised
   """
+  alias ElixirAnalyzer.Constants
 
   defmacro __using__(_opts) do
     quote do
-      feature "raises function clause error" do
+      feature Constants.solution_raise_fn_clause_error() do
```
jiegillet, line 9: This is a correction that has nothing to do with this PR. We took the convention of common checks names matching the comment so that they can be used to suppress other checks. I missed in in the PR review.

test/support/analyzer_verification/suppress_if.ex
```diff
@@ -18,10 +18,7 @@ defmodule ElixirAnalyzer.Support.AnalyzerVerification.SuppressIf do
 
   assert_no_call "assert 1: no foo() unless common check was found" do
     comment "assert 1: foo() was called"
-    # This should work:
-    # suppress_if Constants.solution_debug_functions(), :fail
-    # but in assert_call, it gets treated as an AST, so the final string is required 
-    suppress_if "elixir.solution.debug_functions", :fail
+    suppress_if Constants.solution_debug_functions(), :fail
```
jiegillet, line 21: So satisfying 😄

angelikatyborska (approved):
If this works, then it's good enough for me 😁 looks good.

Issue #164, opened 2021-09-27
Inaccurate analyzer comment in `newsletter`


jiegillet:
<img width="1234" alt="Screen Shot 2021-09-27 at 10 49 29" src="https://user-images.githubusercontent.com/16929078/134834238-60784665-03b6-4302-9fb4-a899a562bd8e.png">

Even though I wrote this analyzer check, I still got it wrong :)

I used `IO.write` instead of `IO.puts`. In hindsight, using `IO.puts` is better, but it's not mentioned in the instructions or the hints. I got there because I rushed a bit (like always) and started with `File.write`, then swapped `File` to `IO`.

I can think of a couple of ways to fix this:

1. Mention `IO.puts` somewhere in the description/hints
2. Allow `IO.write` in the check and reword the comment to include the possibility
3. Add a check that tells the user "This works, but IO.puts is better because it includes a newline" (and suppress the "ends with `:ok`" check).

I like 3, because that's what I would do as a mentor.

PR #167, opened 2021-09-28
`newsletter` analysis: suggest IO.puts over IO.write

angelikatyborska:
Closes https://github.com/exercism/elixir-analyzer/issues/164

Website copy PR https://github.com/exercism/website-copy/pull/2086

### DIFF
--- a/lib/elixir_analyzer/test_suite/newsletter.ex
+++ b/lib/elixir_analyzer/test_suite/newsletter.ex
@@ -1,6 +1,6 @@
 defmodule ElixirAnalyzer.TestSuite.Newsletter do
   @moduledoc """
-  This is an exercise analyzer extension module for the concept exercise Newsletter 
+  This is an exercise analyzer extension module for the concept exercise Newsletter
   """
 
   alias ElixirAnalyzer.Constants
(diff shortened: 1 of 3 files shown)


lib/elixir_analyzer/test_suite/newsletter.ex
```diff
@@ -20,9 +20,24 @@ defmodule ElixirAnalyzer.TestSuite.Newsletter do
...
   end
 
+  feature "use IO.puts in log_sent_email rather than IO.write" do
+    find :none
+    type :actionable
+    comment Constants.newsletter_log_sent_email_prefer_io_puts()
+
+    form do
+      def log_sent_email(_ignore, _ignore) do
+        _block_ends_with do
```
jiegillet, line 30: I think we don't need `_block_ends_with` because we would miss a solution like
```
ok =  IO.write(file, line <> "\n")
ok
```
Maybe `assert_no_call` is more appropriate for this particulate check.

angelikatyborska: Hmm. I did that on purpose so that solutions that use `IO.write` and don't end with `IO.write` will be pointed to this comment instead which also mentions `IO.puts`: https://github.com/exercism/website-copy/blob/main/analyzer-comments/elixir/newsletter/log_sent_email_returns_implicitly.md

I could add a separate check that will trigger both comments in this case (this one needs to stay like that to correctly suppress the no implicit return check)

jiegillet: In that case, we would be back to the misleading error that lead me to open the issue. 
Keeping this particular check focused on puts vs. write feels better to me. 

How about removing the `suppress_if`? In this case no matter what, if they use `IO.write`, they get 2 comments: one saying use `IO.puts` and one saying put `IO.puts` at the end. Reading those 2 in a row would be fine.

jiegillet: Actually I've re-read your suggestion and yes, it's probably better than mine.

angelikatyborska: Done: [`68b7f74` (#167)](https://github.com/exercism/elixir-analyzer/pull/167/commits/68b7f74ca525239021fb41969e26275d44a6a157)

jiegillet (approved):
Perfect 🎉

PR #171, opened 2021-09-29
Add common check: compiler warnings

### LINKED ISSUE #165: Report compilation warnings as actionable
Dealing with compilation warnings is crucial for writing clean and maintainable Elixir code, and can be a teaching experience for new developers. However, it shouldn't be a blocker preventing people from submitting their solutions. For this reason, it doesn't belong in the test runner, but here.

Unfortunately that means we need to add a compilation step for the student's submission, which will slow down the analyzer runs.

jiegillet:
Closes #165.
Here is what I came up with for checking compiler warnings. 

I didn't want to compile the solution outside of the app, write stderr to a file and then read the file back in, so I opted for using `Code.compile_quoted(code_ast)` inside of a `ExUnit.CaptureIO.capture_io` function. 

It works, but there are many caveats:
* So many tests do not compile or have compiler warnings, I had to ignore all compile errors and filter out all comments. The only tests that do check the feature are from `test/elixir_analyzer_test.exs`.
* I have to purge the compiled modules after each test, otherwise they stay in memory and produce even more warnings ("redefining module XXXXX (current version defined in memory)"). I ran into some problems because some tests used the module name `Module` and everything broke once I purged that (obviously). Meaning a maliciously named module could potentially make the analyzer crash.
* As you guessed, it's a lot slower. I made a basic benchmark (not committed but copied below with results). Running the full test suite goes from 5 seconds to 15 second. 

I'm sure you have suggestions to improve this, but I'm more and more inclined to think that in the long run this should be part of the test runner, and that the platform should add a "compiler warning" field to the test results or something...

For reference, here is `bin/benchmarks.exs`:
```elixir
Logger.configure(level: :error)

options = [puts_summary: false, write_results: false]
perfect_path = "./test_data/two_fer/perfect_solution/"
imperfect_path = "./test_data/two_fer/imperfect_solution/"

Benchee.run(
  %{  
    "perfect two-fer" => fn ->
      ElixirAnalyzer.analyze_exercise("two-fer", perfect_path, perfect_path, options)
    end,
    "imperfect two-fer" => fn ->
      ElixirAnalyzer.analyze_exercise("two-fer", imperfect_path, imperfect_path, options)
    end 
  },  
  time: 10, 
  memory_time: 2
)
```
Ans the benchmarks (cropped a bit):
```
➜  elixir-analyzer git:(jie-compiler-warnings) ✗ mix run bin/benchmarks.exs
Operating System: macOS
CPU Information: Intel(R) Core(TM) i5-5257U CPU @ 2.70GHz
Number of Available Cores: 4
Available memory: 8 GB
Elixir 1.12.0
Erlang 24.0.1

Benchmark suite executing with the following configuration:
warmup: 2 s
time: 10 s
memory time: 2 s
parallel: 1
inputs: none specified
Estimated total run time: 28 s

Benchmarking imperfect two-fer...
Benchmarking perfect two-fer...

Name                        ips        average  deviation         median         99th %
perfect two-fer           63.87       15.66 ms    ±23.42%       15.06 ms       25.67 ms
imperfect two-fer         32.07       31.18 ms    ±16.47%       31.68 ms       49.20 ms

Comparison: 
perfect two-fer           63.87
imperfect two-fer         32.07 - 1.99x slower +15.53 ms

Memory usage statistics:

Name                      average  deviation         median         99th %
perfect two-fer         646.10 KB     ±0.01%      646.13 KB      646.13 KB
imperfect two-fer       747.91 KB     ±0.01%      747.94 KB      747.94 KB

Comparison: 
perfect two-fer         646.13 KB
imperfect two-fer       747.91 KB - 1.16x memory usage +101.81 KB
➜  elixir-analyzer git:(jie-compiler-warnings) ✗ git checkout main
Switched to branch 'main'
Your branch is up to date with 'origin/main'.
➜  elixir-analyzer git:(main) ✗ mix run bin/benchmarks.exs

[...same stuff than above...]

Name                        ips        average  deviation         median         99th %
perfect two-fer          1.08 K        0.93 ms    ±64.37%        0.74 ms        3.46 ms
imperfect two-fer        0.91 K        1.10 ms    ±83.28%        0.79 ms        5.53 ms

Comparison: 
perfect two-fer          1.08 K
imperfect two-fer        0.91 K - 1.19x slower +0.176 ms

Memory usage statistics:

Name                 Memory usage
perfect two-fer         338.73 KB
imperfect two-fer       384.84 KB - 1.14x memory usage +46.10 KB

**All measurements for memory usage were the same**
```


jiegillet:
Interesting, the tests fail because of color output in the console. Not sure how to fix that...

```
 left:  "...{\"comment\":\"elixir.solution.compiler_warnings\", \"params\": {\"warnings\":\"warning: module attribute @someUnusedModuleAttribute was set but never used\\n  nofile:2\\n\\n\"}, \"type\":\"actionable\"}..."
 right: "...{\"comment\":\"elixir.solution.compiler_warnings\", \"params\": {\"warnings\":\"\\u001B[33mwarning: \\u001B[0mmodule attribute @someUnusedModuleAttribute was set but never used\\n  nofile:2\\n\\n\"} ,\"type\":\"actionable\"}...
```

test/elixir_analyzer/exercise_test/common_checks_test.exs
```diff
@@ -187,22 +187,22 @@ defmodule ElixirAnalyzer.ExerciseTest.CommonChecksTest do
     test_exercise_analysis "doesn't report snake_case variable names",
       comments_exclude: [Constants.solution_variable_name_snake_case()] do
       [
-        defmodule Module do
+        defmodule MyModule do
```
angelikatyborska, line 190: Oops, accidental monkey patching 😂

test/support/exercise_test_case.ex
```diff
@@ -97,6 +99,8 @@ defmodule ElixirAnalyzer.ExerciseTestCase do
           comments =
             result.comments
             |> Enum.map(fn comment_details -> comment_details.comment end)
+            # There are too many compiler warnings in tests
```
angelikatyborska, line 102: Good choice. We wouldn't want to make all test snippets perfect, and we also don't need to do any assertions about when which warning is produced because that's Elixir internals.

angelikatyborska:
This happens when `mix` doesn't know that it outputs to an interactive shell. I have no idea what kind of system env vars govern this, but you can reproduce the failure locally with:

```
mix test test/elixir_analyzer_test.exs | cat  
```

Reading through `IO.ANSI` source I found a hint that we could just force the colors, and maybe we need to do that in the whole app, not just the tests / on CI, because in prod it also might run without an interactive shell.

```elixir
Application.put_env(:elixir, :ansi_enabled, true)
```

angelikatyborska (approved):

angelikatyborska:
> imperfect two-fer - 99th % 49.20 ms

The slowest runs for an imperfect two-fer solution with a compilation don't even take 100ms? That doesn't sound like a performance penalty at all 🤔

jiegillet:
lol, I'm sorry I clicked  "Edit" on your last comment instead of "Quote". Your first sentence is lost...

> The slowest runs for an imperfect two-fer solution with a compilation don't even take 100ms? That doesn't sound like a 
 performance penalty at all 🤔 maybe the slowest part of `mix compile` is the file IO, which here we don't add since we compile the AST from memory?

I'm not sure, my gut is telling me that `mix compile` would be slower, but I haven't compared.

test/support/exercise_test_case.ex
jiegillet: Yeah, that would be so much work. Not worth it ^^

jiegillet:
I ended up only disabling ANSI colors around the compiler check. It's a little clunky, but disabling it everywhere meant disabling all colors for the normal tests and I don't want to give that up.

angelikatyborska:
Yay! Let's merge and see what happens. I'm excited for this feature 🤩

PR #173, opened 2021-10-01
Add check for Wine Cellar

### LINKED ISSUE #172: Handle common problems in `wine-cellar`
After browsing a few community solutions I only see one problem here so far. I also had it in my both mentoring sessions for this exercise 🤔 

## Problem 1: not using `Keyword.get_values` for filtering by color

`Keyword.get_values` is by far the shortest and easiest way to fulfill the requirement of filtering by color.

It should be favored over other approaches, also to encourage exploration of the standard library. It's already included in the hints.

Recursive approach
```elixir
 def filter(cellar, color, opts \\ []) do
    cellar
    |> filter_by_color(color)
    |> optional_filter_by_year(Keyword.get(opts, :year))
    |> optional_filter_by_country(Keyword.get(opts, :country))
  end

  defp filter_by_color([], _), do: []
  defp(filter_by_color([{color, wine} | tail], color)) do
    [wine | filter_by_color(tail, color)]
  end

  defp filter_by_color([{_, _} | tail], color) do
    filter_by_color(tail, color)
  end
```

List comprehensions
```elixir
  def filter(cellar, color, opts \\ []) do
    wines = 
      for {^color, wine} <- cellar do
        wine
      end
    opts
    |> Enum.reduce(wines, fn
      {:year, year}, wines -> filter_by_year(wines, year)
      {:country, country}, wines -> filter_by_country(wines, country)
    end)
  end
```

Br1ght0ne:
Assert that the solution uses Keyword.get_values/2.

Depends on https://github.com/exercism/website-copy/pull/2091. Closes #172.

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -81,6 +81,9 @@ config :elixir_analyzer,
     "take-a-number" => %{
       analyzer_module: ElixirAnalyzer.TestSuite.TakeANumber
     },
+    "wine-cellar" => %{
+      analyzer_module: ElixirAnalyzer.TestSuite.WineCellar
+    },
 
     # practice exercises
     "accumulate" => %{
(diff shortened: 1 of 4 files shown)


angelikatyborska (approved):
Perfect!

PR #174, opened 2021-10-01
Add common check for prepending items to lists

### LINKED ISSUE #170: Add common check to suggest using `[h | t]` instead of `[h] ++ t`
@jiegillet writes:

> I think `[element]++list` is kind of bad because
> 1. It’s longer
> 2. It’s harder to read because it is misleading (looks like you append actual lists, not cons an element)
> 3. It takes longer to run (probably)
> 4. It is a hint that the person who wrote that doesn’t know  about `[element | list]`

Br1ght0ne:
Assert that `[h | t]` is used instead of `[h] ++ t`. Also replace `[h] ++ t` syntax in Accumulate test :)

Depends on https://github.com/exercism/website-copy/pull/2092. Closes #170.

### DIFF
--- a/test/elixir_analyzer/exercise_test/common_checks/list_prepend_head_test.exs
+++ b/test/elixir_analyzer/exercise_test/common_checks/list_prepend_head_test.exs
@@ -1,0 +1,36 @@
+defmodule ElixirAnalyzer.Support.AnalyzerVerification.ListPrependHead do
+  use ElixirAnalyzer.ExerciseTest
+end
+
+defmodule ElixirAnalyzer.ExerciseTest.CommonChecks.ListPrependHeadTest do
+  use ElixirAnalyzer.ExerciseTestCase,
+    exercise_test_module: ElixirAnalyzer.Support.AnalyzerVerification.ListPrependHead
+
+  alias ElixirAnalyzer.Constants
+
+  test_exercise_analysis "Solutions using [h | t] to prepend to list",
+    comments: [] do
+    [
+      defmodule MyModule do
+        def prepend_ok(list) do
+          [:ok | list]
+        end
+
+        def concat_lists() do
+          [1, 2, 3] ++ [4, 5, 6]
+        end
+      end
+    ]
+  end
+
+  test_exercise_analysis "Solutions using [h] ++ t to prepend to list",
+    comments: [Constants.solution_list_prepend_head()] do
+    [
+      defmodule MyModule do
+        def prepend_ok(list) do
+          [:ok] ++ list
+        end
+      end
+    ]
+  end
+end
--- a/test/elixir_analyzer/exercise_test/common_checks/list_prepend_head_test.exs
+++ b/test/elixir_analyzer/exercise_test/common_checks/list_prepend_head_test.exs
@@ -15,9 +15,15 @@ defmodule ElixirAnalyzer.ExerciseTest.CommonChecks.ListPrependHeadTest do
         def prepend_ok(list) do
           [:ok | list]
         end
-
-        def concat_lists() do
+      end,
+      defmodule MyModule do
+        def concat_equal_lists() do
           [1, 2, 3] ++ [4, 5, 6]
+        end
+      end,
+      defmodule MyModule do
+        def append_element() do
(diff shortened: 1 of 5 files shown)


test/elixir_analyzer/exercise_test/common_checks/list_prepend_head_test.exs
```diff
@@ -0,0 +1,36 @@
...
+      end
+    ]
+  end
+
+  test_exercise_analysis "Solutions using [h] ++ t to prepend to list",
+    comments: [Constants.solution_list_prepend_head()] do
+    [
+      defmodule MyModule do
+        def prepend_ok(list) do
+          [:ok] ++ list
```
jiegillet, line 31: Just for piece of mind, could you add a couple more complex examples? Adding a head to a function that produces a list, a prepend inside of a pipe, things like that.

test/elixir_analyzer/test_suite/accumulate_test.exs
```diff
@@ -8,7 +8,7 @@ defmodule ElixirAnalyzer.ExerciseTest.AccumulateTest do
       def accumulate([], _), do: []
 
       def accumulate([head | tail], fun) do
-        [fun.(head)] ++ accumulate(tail, fun)
+        [fun.(head) | accumulate(tail, fun)]
```
jiegillet, line 11: That’s the exemplar solution I think… @angelikatyborska should we update it on the elixir repo as well?

jiegillet (approved):
Looks great, thank you! I left a couple of comments. 
are you participating to hacktoberfest? We can tag this PR accordingly

test/elixir_analyzer/test_suite/accumulate_test.exs
angelikatyborska: I'm not seeing anything that needs updating in the Elixir repo, it already uses `|`: https://github.com/exercism/elixir/blob/main/exercises/practice/accumulate/.meta/example.ex#L24

Anyway, remember about the difference between concept exercises and practice exercises. Only  concept exercises have _exemplar_ solutions. Practice exercises have _example_ solutions which can be less than ideal as long as they work 😁

jiegillet: Oh I'm sorry, I just assumed. I didn't check because I was reviewing from my phone. I'll stop doing that, I think ^^

I know the difference between exemplar and example, I just didn't really remember which type `accumulate` was 😅
I should use a generic word... Maybe "exemploid"

angelikatyborska: It's ok, don't top! If you can check yourself, it's better to check before complaining, but if you can't, it's better to complain and be wrong than not complain and let a problem slip. At least in my book.

I like "exemploid" :D sooner or later other maintainers will have the same issue with vocabulary I guess, how to talk about both exemplars and examples and be clear that you mean both...

jiegillet: You really have a way with making people feel better about their mistakes, have I told you that? 💯 

And about  exemploids, careful of the [standard proliferation](https://xkcd.com/927/) ;)

Br1ght0ne: Anything to act on here? Just lurking around your conversation 👀

test/elixir_analyzer/exercise_test/common_checks/list_prepend_head_test.exs
Br1ght0ne: Added in 2bb23cd. Thanks!

test/elixir_analyzer/test_suite/accumulate_test.exs
jiegillet: Haha, sorry about that!
Nope, I think we are good, my initial comment was mistaken.

jiegillet:
One more dose of `mix format` and we should be good to go!

angelikatyborska:
I don't really see a need for `mix format` but it was the comment check that needed to be rerun 🙂. Merging!

jiegillet:
Ah, I think I was thinking of #162.

PR #180, opened 2021-10-02
Add a common check for boolean function naming patterns

### LINKED ISSUE #168: Add a common check for `is_foo?` function names that mix up two separate naming patterns
I saw this in one of the community solutions:

```
defp is_allowed_char?(char), do: (char >= 97 and char <= 122) or char == 95
```

The correct name is either `allowed_char?` or `is_allowed_char`, not both. Source: https://hexdocs.pm/elixir/naming-conventions.html#is_-prefix-is_foo

We should add a common check that looks for functions and macros named `is_foo?` and points to the explanation that it should be either `foo?` or `is_foo` (but we don't need to try to detect which it should be).

jiegillet:
Closes #168.

This one is faulty straightforward.

### DIFF
--- a/lib/elixir_analyzer/exercise_test/common_checks/boolean_functions.ex
+++ b/lib/elixir_analyzer/exercise_test/common_checks/boolean_functions.ex
@@ -1,0 +1,86 @@
+defmodule ElixirAnalyzer.ExerciseTest.CommonChecks.BooleanFunctions do
+  @moduledoc """
+  Reports the first boolean function with a name that's not appropriate.
+
+  Doesn't report more if there are more.
+  A single comment should be enough for the student to know what to fix.
+
+  Common check to be run on every single solution.
+  """
+
+  alias ElixirAnalyzer.Constants
+  alias ElixirAnalyzer.Comment
+
+  @spec run(Macro.t()) :: [{:pass | :fail | :skip, %Comment{}}]
+  def run(ast) do
+    {_, %{function_names: function_names}} =
+      Macro.prewalk(ast, %{function_names: []}, &traverse/2)
+
+    case List.last(function_names) do
+      nil ->
+        []
+
+      {def_type, name} ->
+        {wrong_def, correct_def} = render_names(def_type, name)
+
+        [
+          {:fail,
+           %Comment{
+             type: :actionable,
+             name: Constants.solution_boolean_functions(),
+             comment: Constants.solution_boolean_functions(),
+             params: %{
+               expected: correct_def,
+               actual: wrong_def
+             }
+           }}
+        ]
+    end
+  end
+
+  @def_ops [:def, :defp, :defmacro, :defmacrop, :defguard, :defguardp]
+  defp traverse({op, _meta, [{name, _meta2, _parameters} | _]} = ast, acc) when op in @def_ops do
+    name = to_string(name)
+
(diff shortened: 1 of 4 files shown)


lib/elixir_analyzer/exercise_test/common_checks/boolean_functions.ex
```diff
@@ -0,0 +1,85 @@
...
+    cond do
+      starts_with_is and end_with_? -> false
+      starts_with_is and type in [:def, :defp] -> false
+      end_with_? and type in [:defguard, :defguardp] -> false
+      true -> true
+    end
+  end
+
+  defp render_names(type, name) do
+    is_encloded_? = String.starts_with?(name, "is_") and String.ends_with?(name, "?")
```
angelikatyborska, line 69: Encloded - is that a word I don't know yet or a typo?

lib/elixir_analyzer/exercise_test/common_checks/boolean_functions.ex
```diff
@@ -0,0 +1,85 @@
...
+    correct_name =
+      cond do
+        type in [:def, :defp] and is_encloded_? -> name_without_is
+        type in [:def, :defp] -> name_without_is <> "?"
+        type in [:defguard, :defguardp] and is_encloded_? -> name_without_?
+        type in [:defguard, :defguardp] -> "is_" <> name_without_?
+        true -> "#{name_without_?} or #{type} #{name_without_is}"
+      end
+
+    {"#{type} #{name}", "#{type} #{correct_name}"}
```
angelikatyborska, line 83: I'm wondering about the comment here. In case of `defmacro is_foo?`, we will have those variables.

```
actual = "defmacro is_foo?"
expected = "defmacro is_foo or defmacro foo?"
```

Now we have no way of enclosing both potential names separately with backticks in the markdown comment because they're in a single variable. The best we can have is something like this, which isn't technically correct Elixir code:

> expected:
> ```elixir
> defmacro is_foo or defmacro foo?
> ```

I'm thinking maybe we need three different comments. One for functions and one for guards, both of which have a single expected name, and a third one for macros, that will have two expected names. This would also allow us to write more tailored explanations.

lib/elixir_analyzer/exercise_test/common_checks/boolean_functions.ex
jiegillet: Oh yeah, it's totally a new word that the cool people know about, but for noobs like you, I guess I should change it to `is_enclosed_?` or something.

lib/elixir_analyzer/exercise_test/common_checks/boolean_functions.ex
jiegillet: One way to make it correct code is to change to `# or defmacro foo?`, but your suggestion is better :)

lib/elixir_analyzer/exercise_test/common_checks/boolean_functions.ex
angelikatyborska: As a noob, I am thankful for the change 🙏 After turning 25 I stopped being able to catch up with what's cool now and what isn't. I still try to spell words correctly, like a nerd 😞

angelikatyborska (approved):
Let's goooo!

PR #183, opened 2021-10-03
Add a celebratory summary text

angelikatyborska:
An emoji-only response looks a bit broken. We need a few words.


jiegillet (approved):
Clean and simple, I like it! You're doing something right. 🎉

PR #185, opened 2021-10-04
Add a common check for `@doc` + `@spec` order

### LINKED ISSUE #181: Add common check for `@doc` + `@spec` order
Write a common check that will detect this order of annotations
```elixir
@spec x()
@doc ""
def x()
```
And recommend to do this instead:
```elixir
@doc ""
@spec x()
def x()
```

The type of this comment should be `:actionable` or `:informative`, not `:essential`.

This will mostly apply just to the `city-office` exercise, but in theory it's a common check. I was asked about this in a few mentoring sessions.

While technically it doesn't matter which one comes first, `@spec` or `@doc`, almost always `@doc` comes first. My personal interpretation is that it's because `@spec` depends on the function name and argument list, so that when it's as close to the function definition as possible, it's easier to check if it's correct and to later change both when the function changes.

I don't know if there are any more authoritative opinions on this anywhere that could give better reasons?

Make sure to test this well, with scenarios such as this one not triggering the error:
```elixir
@doc ""
def x

@spec y
def y
```

And remember about other keywords: `defmacro`, `defguard`, and their private equivalents.

joseemds:
This is a initial implementation, i still have to deal with some corner cases where `Enum.chunk_every` will not fit, but first i would like to receive some initial feedback if this implementation is a good beginning, and if it should report just the first case, or report all  cases found.

Closes #181


joseemds:
I added more tests and changed the implementation of `DocOrderSpec.run`, before it was using `Enum.chunk_every(3)` and now its using  `Enum.chunk_by(&(&1 in @def_ops))`.
Before the checker wouldnt work in a piece of code like this (Even if its wrong):
```elixir
 
      defmodule Test do
        def a
        def b

        @doc ""
        @spec c
        def c
      end
```
Because this would be turned into `[[:def, :def, :doc], [:spec, :def]`, and wouldnt match the old `[:doc, :spec, op]` clause.

Now it is chucking every time that other definition is found, with this that piece would turn into this: `[[:def, :def], [:doc, :spec], [:def]]`, it is still not the most readable option, but it makes easier to separate definitions and module attributes, making easier to match the wrong order.

Please let me know if there is a better approach for this or if i should add more tests

jiegillet (commented):
This is a very preliminary review, I haven't read everything in great detail yet, or ran the tests, but I still have some comments.

I feel that this check could get complex quickly if we want to consider all possibilities, so maybe we need to aim for something useful even if imperfect, that avoids false positives.

joseemds:
Hi @jiegillet thanks for your review! I added the check for different function and spec name, and also added params to both errors, i added they thinking that the template would be something like: 

```md
Wrong order: 

@spec @%{fn_name}
@doc ""
@%{@op} @%{fn_name}

Right order: 
@doc ""
@spec @%{fn_name}
@%{@op} @%{fn_name}
```
```
Spec and Function name doesnt match:
@spec @%{spec_name}
@%{@op} @%{fn_name}

It should be:
@spec @%{fn_name}
@%{@op} @%{fn_name}
```

But im not sure if this is the best way of doing it, please let me know if there is something to improve in the code or in the params errors!

jiegillet:
The `@` in the templates are not special characters, the string interpolation symbol is just `%{}` (I got it wrong my first time too :p ). We also prefer that the code in that code block be "actual" elixir code so things like "Right order:" should be commented.

If I were to do it, I think I would go for something simpler like this:
```
# recommended order
%{correct}

# your order
%{actual}
```
where `correct` and `actual` would be the only 2 parameters passed around and they would be created in the code. For example, `correct` could be the string `@doc ""\n@spec function_name\ndefp function_name`, etc.

joseemds:
I changed the `params` to be `correct` and `actual` and will open a PR to the website-copy repo adding the markdown files soon. Let me know if there is something i can improve in the actual implementation or tests.

joseemds:
So i renamed, changed the name and add some more tests, i think that it is pretty solid now, but im not sure if the error used should be the same for `@doc` and `@spec` order errors and for `@doc` or `@spec` between or after definition, what do you think?

neenjaw:
@joseemds, I agree with you

joseemds:
@neenjaw I fixed merge conflicts and some errors, and renamed the files in the website-copy PR, i think this is done, please let me know if i should change something

test/elixir_analyzer/exercise_test/common_checks/function_annotation_order_test.exs
```diff
@@ -0,0 +1,573 @@
...
+        defmodule Test.Y do
+          @doc ""
+          def x(), do: 1
+        end
+      end
+    end
+  ]
+
+  for {ast, n} <- Enum.with_index(asts) do
+    test "neenjaw -- false positive test ##{n}" do
```
neenjaw, line 569:
```suggestion
    test "##{n}: sub-modules should not raise false positive error" do
```

test/elixir_analyzer/exercise_test/common_checks/function_annotation_order_test.exs
```diff
@@ -0,0 +1,573 @@
...
+          @doc ""
+          def x
+
+          @spec y
+          def y
+        end
+      end
+
+    assert FunctionAnnotationOrder.run(ast) == []
+  end
```
neenjaw, line 143: Can we also test this? I didn't see it in the suite, but if it already exists, feel free to disregard:

```elixir
  test "another non related definition will not make crash" do
    ast =
      quote do
        defmodule Test do
          @spec x
          def x

          @doc ""
          def y
        end
      end

    assert FunctionAnnotationOrder.run(ast) == []
  end
```

neenjaw (approved):
@joseemds, nice work, I think this is pretty much ready.

I have two minor suggestions, feel free to use or ignore.

Next steps:
- Creating a corresponding analyzer comment in the exercism/website-copy repo
  - once this is created and merged, the final CI check will pass
- @angelikatyborska is still required to approve as she is a code-owner in this repo
- @jiegillet may or may not have follow up comments as well

Thanks for sticking it through.  Props to you!

angelikatyborska:
I feel weird as the only one with merge rights 🙂. I didn't involve myself in this PR because I think 2 reviewers is already enough, so I am happy to "blindly" approve and merge when @jiegillet too agrees it's ready (and @jiegillet can you also take care of merging the comment https://github.com/exercism/website-copy/pull/2098/files ?)

jiegillet (commented):
Thank you for you work on this difficult task!

I do have one comment before approving. @angelikatyborska I know you didn't follow the PR closely, but this is about communication with the students so I would like to have your opinion.

test/elixir_analyzer/exercise_test/common_checks/function_annotation_order_test.exs
```diff
@@ -0,0 +1,573 @@
...
+                  expected: """
+                  @doc
+                  @spec test
+                  def test
+                  """,
+                  actual: """
+                  @spec test
+                  @doc
+                  def test
+                  """
```
jiegillet, line 444: I have an issue here. In many of the failing tests, the actual code has little to do with what we say is `actual`. In this particular example, there is no `@doc` at all, in the one below there is no `@spec`. This is guaranteed to confuse the students. 
Having an accurate `actual` is probably hard to do since the full picture is often broken between chunks, so my proposal is to remove the field altogether. We don't tell what the student did wrong specifically, we simply show them what is expected. Since they have the function name they will be able to figure it out. It will be a bit more work, but I believe leaving things as they are will be counter productive.
What do you think?

joseemds: I have been thinking about it, and actually it is not that hard, the chunks are per function name now, for example:
```elixir
        defmodule Test do
          def a(x \\ [])
          @doc ""
          @spec a(list()) :: atom()
          def a([]), do: :empty
          def a(_), do: :full

          @spec b
          def b

          @spec c
          def c
        end
```

Will generate the following chunk to `a` functions and attrs: `[:def, :doc, :spec, :def, :def]` So we could use this to define the `actual` field in a more precise way, and also define a sort function to use the sorted order in the `expected` field, i have not defined an algorithm to sort it yet, but i think it is not that hard.

What do you think about this?

jiegillet: Sorting should be easy, but code like this
```elixir 
def a
@spec a
@doc ""
def a
```
Will be split into 2 different chunks right? `[:def, :spec, :doc]` and `[:def]`, so in general it might be a bit tricky to get the real full picture if you need to look at the previous/next chunk. Feel free to give it a go, but I'm wondering if it's worth the trouble.

angelikatyborska: > @angelikatyborska I know you didn't follow the PR closely, but this is about communication with the students so I would like to have your opinion.

IMO we don't need any code blocks in the comment shown to the student. This text should be more than enough to figure out what the check is about:

>  Developers can choose the order of the `@doc` and `@spec` modules attributes, but the Elixir community convention is to use `@doc` first and `@spec` next to the function.

joseemds: Done! Since there is no code snippets, i limited the error to show only once

test/elixir_analyzer/exercise_test/common_checks/function_annotation_order_test.exs
```diff
@@ -0,0 +1,373 @@
+defmodule ElixirAnalyzer.ExerciseTest.CommonChecks.FunctionAnnotationOrderTest do
```
jiegillet, line 1: One last thing to make the tests cleaner. Now that we don't have an error with parameters, I think we can switch the style of tests to make them more readable. Could you change them to [look like these](https://github.com/exercism/elixir-analyzer/blob/main/test/elixir_analyzer/exercise_test/common_checks/unless_with_else.exs).

jiegillet (commented):
Almost done. 
I see there are also a couple of conflicts, you'll need to fix those too.

lib/elixir_analyzer/exercise_test/common_checks.ex
```diff
@@ -47,6 +48,9 @@ defmodule ElixirAnalyzer.ExerciseTest.CommonChecks do
       ModulePascalCase.run(code_ast),
       CompilerWarnings.run(code_path, code_ast),
       BooleanFunctions.run(code_ast),
+      Indentation.run(code_ast, code_string),
+      FunctionAnnotationOrder.run(code_ast),
+      Comments.run(code_ast, code_string),
```
jiegillet, line 53: I think you ended up adding extra `Indentation` and `Comments` functions in the merge, that's why there is a test failing. Those two are repeated below.

test/elixir_analyzer/exercise_test/common_checks/function_annotation_order_test.exs
jiegillet: Very nice, thank you for the effort!

joseemds:
I couldnt undestarnd why the tests failed, if you can point me the reason i can fix it!

jiegillet (commented):
Awesome, it's looking good! Almost ready to merge.
One more thing I commented on, and also, could you update the git submodule to the latest version? 
```bash
git submodule update --init --recursive --remote
```

joseemds:
Done, now the tests are passing!

jiegillet:
Yay! update the submodule and I'm merging :)

joseemds:
done!

jiegillet:
Thank you very much for your hard work over the last 2 months!
@neenjaw, thank you as well for your all your valuable input.

PR #186, opened 2021-10-04
Add high school sweetheart analyzer

### LINKED ISSUE #86: Write analysis for concept exercise `high-school-sweetheart`
Desired checks: https://github.com/exercism/elixir/blob/12acf3405242970cd492590e2c9d503330188364/exercises/concept/high-school-sweetheart/.meta/design.md#L31-L37

Note that it's impossible to tell from the AST if the multiline string syntax was used, so we can't do that check.

Lgdev07:
Closes #86

Since we need more discussion on multiline string, I only implemented the functions reuses analysis.

### DIFF
--- a/lib/elixir_analyzer/constants.ex
+++ b/lib/elixir_analyzer/constants.ex
@@ -71,6 +71,9 @@ defmodule ElixirAnalyzer.Constants do
       "elixir.high-score.use_default_argument_with_module_attribute",
     high_score_use_map_update: "elixir.high-score.use_map_update",
 
+    # High School Sweetheart Comments
+    high_school_sweetheart_function_reuse: "elixir.high_school_sweetheart.function_reuse",
+
     # Language List Comments
     language_list_do_not_use_enum: "elixir.language-list.do_not_use_enum",
 
--- a/lib/elixir_analyzer/test_suite/high_school_sweetheart.ex
+++ b/lib/elixir_analyzer/test_suite/high_school_sweetheart.ex
@@ -1,0 +1,30 @@
+defmodule ElixirAnalyzer.TestSuite.HighSchoolSweetheart do
+  @moduledoc """
+  This is an exercise analyzer extension module for the concept exercise HighSchoolSweetheart
+  """
+
+  alias ElixirAnalyzer.Constants
+
+  use ElixirAnalyzer.ExerciseTest
+
+  assert_call "initial/1 reuses first_letter/1" do
+    type :actionable
+    calling_fn module: HighSchoolSweetheart, name: :initial
+    called_fn name: :first_letter
+    comment Constants.high_school_sweetheart_function_reuse()
+  end
+
+  assert_call "initials/1 reuses initial/1" do
+    type :actionable
+    calling_fn module: HighSchoolSweetheart, name: :initials
+    called_fn name: :initial
+    comment Constants.high_school_sweetheart_function_reuse()
+  end
+
+  assert_call "pair/1 reuses initials/1" do
+    type :actionable
+    calling_fn module: HighSchoolSweetheart, name: :pair
+    called_fn name: :initials
+    comment Constants.high_school_sweetheart_function_reuse()
+  end
+end
--- a/test/elixir_analyzer/test_suite/high_school_sweetheart_test.exs
+++ b/test/elixir_analyzer/test_suite/high_school_sweetheart_test.exs
@@ -1,0 +1,116 @@
+defmodule ElixirAnalyzer.TestSuite.HighSchoolSweetheartTest do
+  alias ElixirAnalyzer.Constants
+
+  use ElixirAnalyzer.ExerciseTestCase,
+    exercise_test_module: ElixirAnalyzer.TestSuite.HighSchoolSweetheart
+
+  test_exercise_analysis "example solution",
+    comments: [] do
+    defmodule HighSchoolSweetheart do
+      def first_letter(name) do
+        name
+        |> String.trim()
+        |> String.first()
+      end
+
+      def initial(name) do
+        name
+        |> first_letter()
+        |> String.upcase()
+        |> Kernel.<>(".")
+      end
+
+      def initials(full_name) do
+        [first_name, last_name] = String.split(full_name)
+        "#{initial(first_name)} #{initial(last_name)}"
+      end
+
+      def pair(full_name1, full_name2) do
+        i1 = initials(full_name1)
+        i2 = initials(full_name2)
+
+        """
+              ******       ******
+            **      **   **      **
+          **         ** **         **
+        **            *            **
+        **                         **
+        **     #{i1}  +  #{i2}     **
+          **                       **
+            **                   **
+              **               **
+                **           **
+                  **       **
+                    **   **
+                      ***
+                      *
+        """
+      end
+    end
+  end
+
+  describe "function reuse" do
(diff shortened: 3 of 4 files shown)


lib/elixir_analyzer/constants.ex
```diff
@@ -71,6 +71,9 @@ defmodule ElixirAnalyzer.Constants do
       "elixir.high-score.use_default_argument_with_module_attribute",
     high_score_use_map_update: "elixir.high-score.use_map_update",
 
+    # High School Sweetheart Comments
+    high_school_sweetheart_function_reuse: "elixir.high_school_sweetheart.function_reuse",
```
jiegillet, line 75: Maybe we should split this comment into 3, because they are about different functions.

lib/elixir_analyzer/test_suite/high_school_sweetheart.ex
```diff
@@ -0,0 +1,30 @@
...
+  end
+
+  assert_call "initials/1 reuses initial/1" do
+    type :actionable
+    calling_fn module: HighSchoolSweetheart, name: :initials
+    called_fn name: :initial
+    comment Constants.high_school_sweetheart_function_reuse()
+  end
+
+  assert_call "pair/1 reuses initials/1" do
```
jiegillet, line 24: It's `pair/2` isn't it?

test/elixir_analyzer/test_suite/high_school_sweetheart_test.exs
```diff
@@ -0,0 +1,116 @@
+defmodule ElixirAnalyzer.TestSuite.HighSchoolSweetheartTest do
+  alias ElixirAnalyzer.Constants
+
+  use ElixirAnalyzer.ExerciseTestCase,
+    exercise_test_module: ElixirAnalyzer.TestSuite.HighSchoolSweetheart
+
+  test_exercise_analysis "example solution",
+    comments: [] do
+    defmodule HighSchoolSweetheart do
```
jiegillet, line 9: Is this the exemplar solution? If yes, since #157 you would need to add one more comment like
> https://github.com/exercism/elixir-analyzer/blob/ca5fd50ff10ba572186690473d13b2befd9ccc41/test/elixir_analyzer/test_suite/bird_count_test.exs#L6

test/elixir_analyzer/test_suite/high_school_sweetheart_test.exs
```diff
@@ -0,0 +1,116 @@
...
+                      *
+        """
+      end
+    end
+  end
+
+  describe "function reuse" do
+    test_exercise_analysis "HighSchoolSweetheart.initial must call first_letter",
+      comments_include: [Constants.high_school_sweetheart_function_reuse()] do
+      defmodule HighSchoolSweetheart do
```
jiegillet, line 55: If you expect the same comments for a bunch of test modules, you can put them together in a list
```elixir
test_exercise_analysis "something",
  comments_include: [Constants.high_school_sweetheart_function_reuse()] do
  [ defmodule HighSchoolSweetheart do
     ...
     end,
     defmodule HighSchoolSweetheart do
     ...
     end
  ]
```

jiegillet: That being said, if we split the comment in three, it can stay this way :)

test/elixir_analyzer/test_suite/high_school_sweetheart_test.exs
```diff
@@ -0,0 +1,116 @@
...
+        end
+      end
+    end
+
+    test_exercise_analysis "HighSchoolSweetheart.pair must call initials",
+      comments_include: [Constants.high_school_sweetheart_function_reuse()] do
+      defmodule HighSchoolSweetheart do
+        def pair(full_name1, full_name2) do
+          i1 = full_name1
+          i2 = full_name2
```
jiegillet, line 94: This solution isn't going to pass the tests... It's not strictly necessary that all test modules pass tests or even compile, but they should reflect something that students might write. The tests are really a justification on why we have those checks in place.

And yes, that sometimes means we have to write contrived code that "no one would ever write, right?", but actually, yes, someone would :)

jiegillet (commented):
Thank you for the PR, it's a great start :)

I left some comments. Make sure you rebase on/merge the most up-to-date version of `main` before running the tests.

Lgdev07:
Thank you very much for the review @jiegillet, I will adjust based on your comments :smile:

lib/elixir_analyzer/constants.ex
Lgdev07: I did write based on this comment about function reuse, maybe we can maintain this one, what do you think?

https://github.com/exercism/elixir-analyzer/blob/ca5fd50ff10ba572186690473d13b2befd9ccc41/lib/elixir_analyzer/constants.ex#L41

test/elixir_analyzer/test_suite/high_school_sweetheart_test.exs
jiegillet: Ah yes, I had a look. I think the problem is since the code is inside of a block, the indentation of the multiline strings is all messed up. There is a fix, you can pass the code as a string in this way:

```elixir
  test_exercise_analysis "example solution",
    comments: [ElixirAnalyzer.Constants.solution_same_as_exemplar()] do
    ~S'''
    defmodule HighSchoolSweetheart do
    ...
    end
    '''
  end
```
If the the input is a string, it will get parsed as an AST straight way anyway, but at least in this way you're in control of the indentation. 
Multiline strings are a pain to deal with, we had several hardships related to that :)

jiegillet:
Nicely done. Make tests pass and that's a wrap :)

test/elixir_analyzer/test_suite/high_school_sweetheart_test.exs
jiegillet: You also need to run `mix format` on this file.

angelikatyborska:
> Make tests pass and that's a wrap :)

Not so fast! We still need a PR in https://github.com/exercism/website-copy with the actual comment 🙂

test/elixir_analyzer/test_suite/high_school_sweetheart_test.exs
Lgdev07: It worked!! thanks, I tried almost everything :laughing:

Lgdev07:
> > Make tests pass and that's a wrap :)
> 
> Not so fast! We still need a PR in https://github.com/exercism/website-copy with the actual comment slightly_smiling_face

Hi @angelikatyborska, I created the PR: https://github.com/exercism/website-copy/pull/2097 :)

test/elixir_analyzer/test_suite/high_school_sweetheart_test.exs
jiegillet: Believe me, I spent hours over weeks on something like this, I even found a bug in the language (that was already patched in Elixir `main` but still), I am now an Elixir multiline pro debugger.

lib/elixir_analyzer/constants.ex
jiegillet, line 75: Tests are failing, you need to  change to
```suggestion
    high_school_sweetheart_function_reuse: "elixir.high-school-sweetheart.function_reuse",
```

jiegillet (approved):
Thank you again, great work :)

angelikatyborska (approved):

Issue #191, opened 2021-10-10
Is it possible to improve the analyzer for `captains-log` ?


jlirochon:
On my [3rd iteration](https://exercism.org/tracks/elixir/exercises/captains-log/solutions/jlirochon) of this exercise, the analyzer gives this advice:

> One purpose of this exercise is to teach how to use Erlang libraries. Format the date using the `:io_lib` module from Erlang.

But as [you can see](https://exercism.org/tracks/elixir/exercises/captains-log/solutions/jlirochon) I used `:io_lib.format/2` already. I think what confuses the analyzer is that I have 2 bodies for `format_stardate/1`.

Is it possible to improve this ? I will happily fix it if you can give me some hints.

By the way, I had 2 bodies for `format_stardate/1` because the tests requires an `ArgumentError` to be raised when you call `format_stardate/1`. As my 1st implementation didn't call `:io_lib.format/2`, this was not the case. I added a `is_float` guard but then a `FunctionClauseError` was raised so I added another body to catch other argument types and raise the required `ArgumentError `.

I get the point now but the overall experience was not as great as it could be. I would like to improve this, if this is reachable for me.

angelikatyborska:
Thank you for the issue! You're right, this sounds like a bad experience that should be improved.

> I think what confuses the analyzer is that I have 2 bodies for format_stardate/1.

Right. After some short debugging, I came to the conclusion that the problem is not multiple clause functions in general, but the fact that the clause that does the desired call to `:io_lib` has a guard expression with `when`.

You can see this bug in action by adding this test case to the file `test/elixir_analyzer/exercise_test/assert_call_test.exs`. This test should pass, but it doesn't:

```elixir
  test_exercise_analysis "works when calling function has a guard expression",
                         comments: [] do
    defmodule AssertCallVerification do
      def function(y) when y > 10 do
        x = List.first([1, 2, 3])
        result = helper()
        IO.puts(result)

        private_helper() |> IO.puts()
      end

      def helper do
        :helped
      end

      defp private_helper do
        :privately_helped
      end
    end
  end
```

The bug is somewhere in the implementation in `lib/elixir_analyzer/exercise_test/assert_call/compiler.ex`. Trying to fix it might be a bit too hard for a first contribution to the project, but you're welcome to try if you feel like it! Otherwise me or @jiegillet would take care of it soon.

jiegillet:
Oh, that's a huge oversight, I'm surprised this went through unnoticed for so long! 
[...]

PR #192, opened 2021-10-10
Resolve exercism/elixir-analyzer#191 : make extract_function_name/1 work when functions use guards

jlirochon:
I wrote some tests but I don't know where they should be placed...

```elixir
defmodule CompilerTest do
  use ExUnit.Case

  test "extract the name of simple functions" do
    quoted = quote do: def(foo(x, y), do: :a)
    assert Compiler.extract_function_name(quoted) == :foo
  end

  test "extract the name of functions using one guard" do
    quoted = quote do: def(foo(x, y) when is_atom(x), do: :a)
    assert Compiler.extract_function_name(quoted) == :foo
  end

  test "extract the name of functions using multiple guards" do
    quoted = quote do: def(foo(x, y) when is_atom(x) when is_integer(y), do: :a)
    assert Compiler.extract_function_name(quoted) == :foo
  end

  test "extract the name of functions using one guard with \"and\" operator" do
    quoted = quote do: def(foo(x, y) when is_atom(x) and is_integer(y), do: :a)
    assert Compiler.extract_function_name(quoted) == :foo
  end
end
```

Mentoring welcome ! Can I do better ?

### DIFF
--- a/test/elixir_analyzer/exercise_test/assert_call/multiple_clause_functions_test.exs
+++ b/test/elixir_analyzer/exercise_test/assert_call/multiple_clause_functions_test.exs
@@ -1,0 +1,144 @@
+defmodule ElixirAnalyzer.ExerciseTest.AssertCall.MultipleClauseFunctionsTest do
+  use ElixirAnalyzer.ExerciseTestCase,
+    exercise_test_module:
+      ElixirAnalyzer.Support.AnalyzerVerification.AssertCall.MultipleClauseFunctions
+
+  test_exercise_analysis "Calling Map.new outside of the observed function",
+    comments: [
+      "didn't find a call to Map.new/0 in function/1"
+    ] do
+    [
+      # Called in another function.
+      defmodule AssertCallVerification do
+        def function(b) do
+          %{a: b}
+        end
+
+        def not_me(b) do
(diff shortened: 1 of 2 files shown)


jlirochon:
Hi @angelikatyborska, I added some tests by following your instructions, thank you !

Not sure if it's good enough. Let me know if you want me to improve something 🙂.

angelikatyborska:
The tests that you've added are beautiful 💜. I second @jiegillet suggestion, a few more tests for `"didn't find a call to Map.new/0 in function/1"` and it will be perfect.

jlirochon:
Fixed credo warnings.

Before adding more tests I would like to have a better understanding of this block copied from sibling files :

```elixir
  assert_no_call "does not call :rand.normal in specific function" do
    type :informative
    calling_fn module: AssertCallVerification, name: :function
    called_fn module: :rand, name: :normal
    comment "found a call to :rand.normal in function/0"
  end
```

I dont understand why the `comment` is the negative of the description. It seems counter-intuitive to me. Conversely if I want this comment to appear when testing some code blocks, why do I need to use `assert_no_call` instead of `assert_call` ? I tried to read the code of these macros but I'm unable to clarify what happens with regard to the comment.

Issue #81, opened 2021-07-07
Write analysis for concept exercise `boutique-inventory`


angelikatyborska:
Desired checks:
https://github.com/exercism/elixir/blob/9774308bb6b898df6ecbde64c5c6e48c651870ae/exercises/concept/boutique-inventory/.meta/design.md#L31-L36

jiegillet:
This one can be done with existing tools.

angelikatyborska:
This issue has been resolved a while ago in: https://github.com/exercism/elixir-analyzer/pull/194

PR #194, opened 2021-10-12
Add analysis to boutique-inventory

fmmatheus:
#81

### DIFF
--- a/test/elixir_analyzer/test_suite/boutique_inventory_test.exs
+++ b/test/elixir_analyzer/test_suite/boutique_inventory_test.exs
@@ -4,51 +4,74 @@ defmodule ElixirAnalyzer.ExerciseTest.BoutiqueInventoryTest do
 
   test_exercise_analysis "example solution",
     comments: [ElixirAnalyzer.Constants.solution_same_as_exemplar()] do
-    [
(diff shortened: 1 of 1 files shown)


angelikatyborska:
I looked through the published community solutions and found those snippets that would fail the analyzer checks. It will help writing the tests that @jiegillet suggested 🙂 

```elixir
  def sort_by_price(inventory) do
    Enum.sort(inventory, &(&1.price <= &2.price))
  end
```
```elixir
  def increase_quantity(item, count) do
    Map.put(item, :quantity_by_size, Enum.into(item[:quantity_by_size], %{}, fn {k,v} -> {k, v + count} end))
  end
```
```elixir
  def total_quantity(item) do
    item[:quantity_by_size]
    |> Map.values()
    |> Enum.sum()
  end
```

To be honest, `increase_quantity` with `Enum.into/3` instead of `Enum.map` + `Enum.into/1` that we wanted is so much better 😱. I had no idea that's possible. I'm not sure what to do about that. Maybe we should remove that check... or even revert it to say "use only `Enum.into/3`" and update the exercise (hints + exemplar)?  Most published solutions used `Enum.map` + `Enum.into`. Unfortunately `assert_call` doesn't support telling apart function arities.

PR #206, opened 2021-10-21
Add div/rem related check to Sieve

### LINKED ISSUE #193: Add check for `sieve`
The description of the problem says 

> Notice that this is a very specific algorithm, and the tests don't check that you've implemented the algorithm, only that you've come up with the correct list of primes. A good first test is to check that you do not use division or remainder operations (div, /, mod or % depending on the language).

I opened more than a dozen random solutions, literally all of them use `rem`, as well as the example solution. For contrast, [here is my solution](https://exercism.org/tracks/elixir/exercises/sieve/solutions/jiegillet).

Understandably, it is inefficient and uncomfortable for an immutable language to implement a true sieve algorithm, but since those are the instructions, we should probably add that check.

I would ban `rem`, `div`, and the modules`Integer`, `Float`, `:math`. Maybe more? :)

Br1ght0ne:
Close #193. Depends on https://github.com/exercism/website-copy/pull/2108.


jiegillet:
Hi @Br1ght0ne, do you need help with this one?

jiegillet (dismissed):

jiegillet:
@angelikatyborska the CI tests are gone, I don't know what I did. Have you seen this before?

angelikatyborska:
I see this happen from time to time and there's never any explanation. I think this might be just GitHub's internal problems?

angelikatyborska:
I just remembered - close and reopen a PR has the power to trigger Github CI

jiegillet (approved):
Yay! That’s a good trick. Tests do pass, I’m merging.

PR #212, opened 2021-10-30
`assert_call` recognizes captured function for direct and indirect calls

jiegillet:
#208 introduced a bug: since captured functions `function/1` are transformed to `{:/, _, [{:function, [], Elixir}, 2]}` they looked like a variable and were therefore not considered. We didn't have any tests for this case.

This PR fixes that, both recognizing the target function in captured notation, or indirectly via helper functions in captured notations.

### DIFF
--- a/lib/elixir_analyzer/exercise_test/assert_call/compiler.ex
+++ b/lib/elixir_analyzer/exercise_test/assert_call/compiler.ex
@@ -405,11 +405,20 @@ defmodule ElixirAnalyzer.ExerciseTest.AssertCall.Compiler do
       when not is_nil(name) and is_list(args) do
     called =
       case function do
-        {:., _, [{:__MODULE__, _, _}, fn_name]} -> {module, fn_name}
-        {:., _, [{:__aliases__, _, fn_module}, fn_name]} -> {fn_module, fn_name}
-        {:|>, _, [_arg, {fn_name, _, atom}]} when is_atom(atom) -> {module, fn_name}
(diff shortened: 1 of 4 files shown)


lib/elixir_analyzer/exercise_test/assert_call/compiler.ex
```diff
@@ -405,6 +408,7 @@ defmodule ElixirAnalyzer.ExerciseTest.AssertCall.Compiler do
         {:., _, [{:__MODULE__, _, _}, fn_name]} -> {module, fn_name}
         {:., _, [{:__aliases__, _, fn_module}, fn_name]} -> {fn_module, fn_name}
         {:|>, _, [_arg, {fn_name, _, atom}]} when is_atom(atom) -> {module, fn_name}
+        {:/, _, [{fn_name, _, atom}, _arity]} when is_atom(atom) -> {module, fn_name}
```
angelikatyborska, line 411: The check in `matching_function_call?` has a guard to check if arity is an integer, but this one doesn't. Is that difference significant or accidental?

angelikatyborska (approved):
Thank you for taking care of this so fast 💜

lib/elixir_analyzer/exercise_test/assert_call/compiler.ex
jiegillet: I don't think the check is actually necessary, as I don't think `:/` will take any other type of argument. I think I left it in the first one for readability, but somehow didn't in this one. I'll add it here too.

PR #215, opened 2021-11-07
Modify `check_source` to take a `%Source{}` argument instead of a string

jiegillet:
I also added an example in the recipes. I admit I didn't muster the motivation to add a chapter about it in the guide, most likely because it wouldn't make sense without also adding a chapter about `assert_call`.

### DIFF
--- a/docs/recipes.md
+++ b/docs/recipes.md
@@ -131,3 +131,19 @@ feature "description" do
   end
 end
 ```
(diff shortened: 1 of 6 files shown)


angelikatyborska (approved):

angelikatyborska:
Makes sense to me!

Issue #214, opened 2021-11-06
 Write analysis for concept exercise `top-secret`


angelikatyborska:
Desired checks: https://github.com/exercism/elixir/blob/main/exercises/concept/top-secret/.meta/design.md#analyzer

PR #217, opened 2021-11-07
Write analysis for concept exercise top-secret

angelikatyborska:
Closes https://github.com/exercism/elixir-analyzer/issues/214

Website copy PR: https://github.com/exercism/website-copy/pull/2122

### DIFF
--- a/lib/elixir_analyzer/test_suite/top_secret.ex
+++ b/lib/elixir_analyzer/test_suite/top_secret.ex
@@ -1,0 +1,22 @@
+defmodule ElixirAnalyzer.TestSuite.TopSecret do
+  @moduledoc """
+  This is an exercise analyzer extension module for the concept exercise Top Secret
+  """
+
+  use ElixirAnalyzer.ExerciseTest
+  alias ElixirAnalyzer.Constants
+
+  assert_call "decode_secret_message/1 uses to_ast/1" do
+    type :essential
+    calling_fn module: TopSecret, name: :decode_secret_message
+    called_fn module: TopSecret, name: :to_ast
+    comment Constants.top_secret_function_reuse()
+  end
+
+  assert_call "decode_secret_message/1 uses decode_secret_message_part/2" do
+    type :essential
+    calling_fn module: TopSecret, name: :decode_secret_message
+    called_fn module: TopSecret, name: :decode_secret_message_part
+    comment Constants.top_secret_function_reuse()
+  end
+end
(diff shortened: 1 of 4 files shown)


lib/elixir_analyzer/test_suite/top_secret.ex
```diff
@@ -0,0 +1,32 @@
...
+    comment Constants.top_secret_function_reuse()
+  end
+
+  feature "references decode_secret_message_part/2 by capturing it" do
+    find :any
+    type :actionable
+    comment Constants.top_secret_function_capture()
+
+    form do
+      &decode_secret_message_part/2
```
jiegillet, line 29: I like this, and I think we could make a common check too.
For sure the function `fn x, y -> foo(x, y) end` could be written as `&foo/2`. The function `&foo(&1, &2)` could probably overwhelmingly be written as `&foo/2`, but there is the edge case of having more ignored arguments. Not sure if there is sometimes a way to be sure of the number of arguments...

test/elixir_analyzer/test_suite/top_secret_test.exs
```diff
@@ -0,0 +1,193 @@
...
+          {_, acc} = Macro.prewalk(ast, [], &decode_secret_message_part/2)
+
+          acc
+          |> Enum.reverse()
+          |> Enum.join("")
+        end
+      end
+    end
+
+    test_exercise_analysis "TopSecret.decode_secret_message/1 must call decode_secret_message_part/2",
```
jiegillet, line 111: I think this one a copy of the one above. I assume you wanted to re-write `decode_secret_message_part/2`.

jiegillet (commented):
One test is not what it says it is.

BTW, `top-secret-analysis` is the coolest branch name ever.

test/elixir_analyzer/test_suite/top_secret_test.exs
angelikatyborska: Oops! Definitely. I had that weird feeling that I forgot something when I was opening the PR 😂

jiegillet (approved):
Nicely done!

lib/elixir_analyzer/test_suite/top_secret.ex
angelikatyborska: The only way to be sure would be to have the function's real arity somehow. But we can definitely make a common check just for `fn -> end`

jiegillet: Oh I think I had a misconception here, I though that `&foo(&1, &2)` meant an anonymous function that can take many arguments but we only consider 2, but the docs say
>The capture operator can also be used to create anonymous functions that expect at least one argument:
> 
> iex> add = &(&1 + &2)
> iex> add.(1, 2)
> 3
> 
> In such cases, using the capture operator is no different than using fn.

In this case, we don't need to know the real arity since we can assume the code works.

PR #219, opened 2021-11-08
Common check: function capture instead of defining a new anonymous function

### LINKED ISSUE #218: Common check: function capture instead of defining a new anonymous function
Add a common check for all exercises that will detect creating unnecessary anonymous functions and suggest capturing an existing function instead.

An anonymous function is unnecessary when the only operation in its body is calling another function with exactly the same arguments in the same order.

Make sure to detect both ways of creating an anonymous function, with the capture operator `&` and with `fn`.

```elixir
# example 1
# unnecessary new anonymous function
fn n, acc -> n + acc end
# could be
&Kernel.+/2

# example 2
# unnecessary new anonymous function
& String.upcase(&1)
# could be
&String.upcase/1
```

See https://github.com/exercism/elixir-analyzer/pull/217#discussion_r744255070
See the existing analysis done for Top Secret that has this check already, but it's exercise-specific 
    - https://github.com/exercism/elixir-analyzer/blob/main/lib/elixir_analyzer/test_suite/top_secret.ex#L23-L31
	- https://github.com/exercism/website-copy/blob/main/analyzer-comments/elixir/top-secret/function_capture.md

For clarity, here are examples when the new anonymous function is necessary:
```elixir
# necessary because of swapped argument order
fn n, acc -> div(acc, n) end

# necessary because of an extra argument
& IO.inspect(&1, charlists: :as_lists)
```

jiegillet:
Closes #218.

I found some exceptions to the rule for functions or notations that cannot use the captured notation `&foo/2`, and I do wonder if there are more. I ran it on every exemploid and found some bugs (and some things to fix, see #218), maybe I should run the check on a bigger codebase? What's a good one for that use?

I'll make the `website-copy` comment tomorrow.

### DIFF
--- a/lib/elixir_analyzer/exercise_test/common_checks.ex
+++ b/lib/elixir_analyzer/exercise_test/common_checks.ex
@@ -13,6 +13,7 @@ defmodule ElixirAnalyzer.ExerciseTest.CommonChecks do
     ExemplarComparison,
     Indentation,
     PrivateHelperFunctions,
+    FunctionCapture,
     Comments
   }
 
@@ -48,6 +49,7 @@ defmodule ElixirAnalyzer.ExerciseTest.CommonChecks do
       ExemplarComparison.run(code_ast, type, exemploid_ast),
       Indentation.run(code_ast, code_string),
       PrivateHelperFunctions.run(code_ast, exemploid_ast),
+      FunctionCapture.run(code_ast),
       Comments.run(code_ast, code_string)
     ]
     |> List.flatten()
--- a/lib/elixir_analyzer/exercise_test/common_checks/function_capture.ex
+++ b/lib/elixir_analyzer/exercise_test/common_checks/function_capture.ex
@@ -31,7 +31,7 @@ defmodule ElixirAnalyzer.ExerciseTest.CommonChecks.FunctionCapture do
     end
   end
 
-  @exceptions [:<<>>, :{}, :|]
+  @exceptions [:<<>>, :{}]
   defp traverse({:&, _, [{name, _, args}]} = node, functions) when name not in @exceptions do
     wrong_use? =
       args
--- a/lib/elixir_analyzer/exercise_test/common_checks/function_capture.ex
+++ b/lib/elixir_analyzer/exercise_test/common_checks/function_capture.ex
@@ -45,6 +45,12 @@ defmodule ElixirAnalyzer.ExerciseTest.CommonChecks.FunctionCapture do
     end
   end
 
+  # fn -> foo end
+  defp traverse({:fn, _, [{:->, _, [[], {name, _, atom}]}]} = node, functions)
+       when is_atom(atom) do
+    {node, [{:fn, name, nil} | functions]}
+  end
+
   defp traverse({:fn, _, [{:->, _, [args, {name, _, args}]}]} = node, functions)
        when name not in @exceptions do
     args = Enum.map(args, fn {var, _, _} -> var end)
@@ -81,11 +87,18 @@ defmodule ElixirAnalyzer.ExerciseTest.CommonChecks.FunctionCapture do
     {wrong, correct}
   end
 
+  defp format_function({:fn, name, nil}) do
+    correct = "&#{name}/0"
+    wrong = "fn -> #{name} end"
+    {wrong, correct}
+  end
+
   defp format_function({:fn, name, args}) do
     name = format_function_name(name)
     correct = "&#{name}/#{length(args)}"
+    space = if Enum.empty?(args), do: "", else: " "
     args = Enum.join(args, ", ")
-    wrong = "fn #{args} -> #{name}(#{args}) end"
+    wrong = "fn #{args}#{space}-> #{name}(#{args}) end"
     {wrong, correct}
   end
 
--- a/lib/elixir_analyzer/exercise_test/common_checks/function_capture.ex
+++ b/lib/elixir_analyzer/exercise_test/common_checks/function_capture.ex
@@ -8,10 +8,14 @@ defmodule ElixirAnalyzer.ExerciseTest.CommonChecks.FunctionCapture do
 
   @spec run(Macro.t()) :: [{:pass | :fail | :skip, %Comment{}}]
   def run(code_ast) do
-    {_, functions} =
-      Macro.prewalk(code_ast, [], fn ast, functions -> traverse(ast, functions) end)
+    acc = %{capture_depth: 0, functions: []}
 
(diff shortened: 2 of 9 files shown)


lib/elixir_analyzer/exercise_test/common_checks/function_capture.ex
```diff
@@ -0,0 +1,102 @@
...
+             params: %{
+               expected: correct_function,
+               actual: wrong_function
+             }
+           }}
+        ]
+    end
+  end
+
+  @exceptions [:<<>>, :{}, :|]
```
angelikatyborska, line 34: Something's weird - removing `:{}, :|` from this list still results in the tests passing

lib/elixir_analyzer/exercise_test/common_checks/function_capture.ex
```diff
@@ -0,0 +1,102 @@
...
+    {wrong, correct}
+  end
+
+  defp format_function({:fn, name, args}) do
+    name = format_function_name(name)
+    correct = "&#{name}/#{length(args)}"
+    args = Enum.join(args, ", ")
+    wrong = "fn #{args} -> #{name}(#{args}) end"
+    {wrong, correct}
+  end
```
angelikatyborska, line 90: When arity is 0, the result has one space too many after `fn`:

```elixir
iex()> ElixirAnalyzer.ExerciseTest.CommonChecks.FunctionCapture.format_function({:fn, :"IO.ANSI.red", []})
{"fn  -> IO.ANSI.red() end", "&IO.ANSI.red/0"}
```

angelikatyborska:
> I do wonder if there are more

I tried to ask myself - what connect those exceptions? They're all special forms. So we can take a look at https://hexdocs.pm/elixir/Kernel.SpecialForms.html#functions

Normal macros are fine (I checked https://hexdocs.pm/elixir/Kernel.html#is_struct/1 for example), they can use the capture notation.

lib/elixir_analyzer/exercise_test/common_checks/function_capture.ex
jiegillet: I shouldn't code too late at night...

For `:{}` I needed to do a test with at least 3 arguments, for 2 arguments the AST looks different.

For `:|` I think I got confused because the following arguments look very similar
```elixir
{:foo, _, [{:a, [], nil}, {:b, [], nil}]} # from fn a, b -> foo(a, b) end
[{:|, _, [{:a, [], Elixir}, {:b, [], Elixir}]}] # from fn a, b -> [a | b]  end
```
The last one is wrapped  in a list, so they are different and I should remove that exception. Thank you!

jiegillet:
> > I do wonder if there are more
> 
> I tried to ask myself - what connect those exceptions? They're all special forms. So we can take a look at https://hexdocs.pm/elixir/Kernel.SpecialForms.html#functions

I looked into these as well, and didn't see anything else.

jiegillet:
I found another edge case I need to address:
```elixir
Enum.map(slices, &Enum.reduce(&1, fn x, acc -> x * acc end))
```
In this case, we cannot suggest to use `&*/2` because it leads to a compiler error: "nested captures via & are not allowed". I guess I need to traverse the AST keeping track of nested captures.

lib/elixir_analyzer/test_suite/top_secret.ex
```diff
@@ -3,9 +3,11 @@ defmodule ElixirAnalyzer.TestSuite.TopSecret do
   This is an exercise analyzer extension module for the concept exercise Top Secret
   """
 
-  use ElixirAnalyzer.ExerciseTest
   alias ElixirAnalyzer.Constants
 
+  use ElixirAnalyzer.ExerciseTest,
+    suppress_tests: [Constants.solution_use_function_capture()]
```
angelikatyborska, line 9: I wouldn't mind just removing this specialized top-secret check in favor of the custom one

test/elixir_analyzer/exercise_test/common_checks/function_capture_test.exs
```diff
@@ -0,0 +1,379 @@
...
+          end
+
+          def legit(input) do
+            input
+            |> Enum.reduce([], fn n, acc -> div(acc, n) end)
+            |> Enum.reduce([], &div(&2, &1))
+            |> Enum.reduce([], fn n, acc -> String.pad_leading(n, acc, "42") end)
+            |> Enum.reduce([], &String.pad_leading(&1, &2, "42"))
+            |> Enum.map(fn x -> String.split(x, ";") end)
+            |> Enum.map(&String.split(&1, ";"))
```
angelikatyborska, line 37: I thought of two more corner cases that already work as expected, but it might be nice to add them to the tests:
```suggestion
            |> Enum.map(&String.split(&1, ";"))
            # behavior modified by using guards
            |> Enum.map(fn x when is_binary(x) -> String.downcase(x) end)
            # behavior modified by using pattern-matching
            |> Enum.map(fn {a, b} = x -> inspect(x) end)
            |> Enum.map(fn x = {a, b} -> inspect(x) end)
```

lib/elixir_analyzer/exercise_test/common_checks.ex
```diff
@@ -47,6 +48,7 @@ defmodule ElixirAnalyzer.ExerciseTest.CommonChecks do
       ExemplarComparison.run(code_ast, type, exemploid_ast),
       Indentation.run(code_ast, code_string),
       PrivateHelperFunctions.run(code_ast, exemploid_ast),
+      FunctionCapture.run(code_ast),
```
angelikatyborska, line 51: It would be great if there was at least one test somewhere that is going to fail if we accidentally remove this line, probably in `common_checks_test.exs`. I noticed that's not the case here, and also for `BooleanFunctions.run(code_ast)` 🙁

angelikatyborska (approved):
Leaving a final review. As far as I can tell, it's working great now!

Approving, but not merging, since I left some comments that suggest doing a few tiny changes (but they're non-blocking).

test/elixir_analyzer/exercise_test/common_checks/function_capture_test.exs
jiegillet: Good idea, they were in my head but not on the page.

lib/elixir_analyzer/exercise_test/common_checks.ex
jiegillet: Done!

jiegillet:
Thanks for the review ❤️

PR #221, opened 2021-11-09
Create dependabot.yml

FaisalAfroz:


angelikatyborska:
Than you for this PR and all the ones I'm about to merge in other repos too 😁. The config looks almost the same as the one we have in the exercism/elixir repo, so I'm confident it will work.

angelikatyborska (approved):

PR #230, opened 2021-11-14
Look for multiline strings in high-school-sweetheart

angelikatyborska:
Closes https://github.com/exercism/elixir-analyzer/issues/86

i went for a very naive implementation that can be easily fooled by documentation, using multiline strings or just using any string that contains the heredoc syntax delimiters, e.g. `~s(""")`.

The module `Code.Formatter` is private and produces insane output that I don't want to parse... I don't want this check to break if we update Elixir versions.

I also gave up on writing a regex that would try to match that the string is inside of the correct function because a) I doubt my regex skills, b) that would produce false-positives when a private helper has the string.

Website copy PR: https://github.com/exercism/website-copy/pull/2130

### DIFF
--- a/test/elixir_analyzer/test_suite/high_school_sweetheart_test.exs
+++ b/test/elixir_analyzer/test_suite/high_school_sweetheart_test.exs
@@ -161,4 +161,185 @@ defmodule ElixirAnalyzer.TestSuite.HighSchoolSweetheartTest do
       ]
     end
   end
+
+  describe "multiline string" do
+    test_exercise_analysis "accepts different ways of writing multiline strings with interpolation",
+      comments_exclude: [Constants.high_school_sweetheart_multiline_string()] do
+      [
+        ~S'''
+        def pair(full_name1, full_name2) do
+          i1 = initials(full_name1)
+          i2 = initials(full_name2)
+          """
+               ******       ******
+             **      **   **      **
+           **         ** **         **
+          **            *            **
+          **                         **
+          **     #{i1}  +  #{i2}     **
+           **                       **
+             **                   **
+               **               **
+                 **           **
+                   **       **
+                     **   **
+                       ***
+                        *
+          """
+        end
+        ''',
+        ~S'''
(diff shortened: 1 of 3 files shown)


test/elixir_analyzer/test_suite/high_school_sweetheart_test.exs
```diff
@@ -161,4 +161,193 @@ defmodule ElixirAnalyzer.TestSuite.HighSchoolSweetheartTest do
...
+
+    test_exercise_analysis "gets fooled by documentation",
+      comments_exclude: [Constants.high_school_sweetheart_multiline_string()] do
+      ~S'''
+      @moduledoc """
+      Exercism exercise solution
+      """
+      @typedoc """
+      I am a rebel and I don't want to accept that the string() type is not a string
+      so I am defining my own string type!
```
jiegillet, line 338: Damn those rebellious students, always fooling our tests 😆

jiegillet (approved):
Looks good to me. 
That's also how I would have implemented it. Our end goal is teaching that multiline strings exists, so as long as the student is using them somehow, I don't have an issue with the test being a little off.

Does 1.13's `string_to_quoted_with_comments/2` also removes the mutliline structure? There is a good chance it does not, because it would affect the document in a big way.

angelikatyborska:
> Does 1.13's string_to_quoted_with_comments/2 also removes the mutliline structure?

yes 😢 

```
iex(5)> single_line
"\"foo\\nbar\\n\"\n"
iex(6)> multi_line
"\"\"\"\nfoo\nbar\n\"\"\"\n"
iex(7)> Code.string_to_quoted_with_comments!(single_line)
{"foo\nbar\n", []}
iex(8)> Code.string_to_quoted_with_comments!(multi_line)
{"foo\nbar\n", []}
iex(9)> Code.string_to_quoted_with_comments!(single_line) == Code.string_to_quoted_with_comments!(multi_line)
true
```

Issue #60, opened 2021-04-24
Write a basic `flatten-array` analysis


angelikatyborska:
The exercise `flatten-array` is meant to practice recursion and pattern-matching on lists. If it's solved using the `List` or `Enum` functions, it's just way too easy. Thus this exercise should use the same kind of analyzer as `bird-count` has - check that `Enum`, `List`, `Stream` or list comprehensions aren't used.

Introduction to writing an analyzer extension: https://github.com/exercism/elixir-analyzer/blob/main/docs/writing-an-analyzer.md

An example of an analyzer extension that detect usages of disallowed modules: https://github.com/exercism/elixir-analyzer/pull/50/files#diff-39ba60165e48ed55ff5dc41ce966786ae8f1c83abf1cc3e2430e0b8b5692f105

angelikatyborska:
If you're picking up this issue, please also consider doing https://github.com/exercism/elixir-analyzer/issues/67 as they will most likely be identical analyzers.

ThiagosLima:
If it is ok, I can work on this issue.

angelikatyborska:
That would be great 💜

ThiagosLima:
Maybe the `Simple Linked List` exercise should have an analysis like this 🤔

angelikatyborska:
With the simple-linked-list exercise it's not so straight-forward, because the functions `from_list` and `to_list` can be implemented with `Enum` or `List` functions. Our example solution does that: https://github.com/exercism/elixir/blob/efd211b80aa350e32cf76242f09d12ec4d49df14/exercises/practice/simple-linked-list/.meta/example.ex#L59-L73

ThiagosLima:
I was wondering that because the exercise want us to implement the linked list abstraction. If we use built-in lists abstractions seems strange. Someone could just use `Enum` to implement the exercise. Ex:

```elixir
def reverse(list), do: list |> to_list() |> Enum.reverse() |> from_list()
```

The functions `from_list` and `to_list` could be implemented using recursion. Ex:

```elixir
def from_list([]), do: {}
def from_list([head | tail]), do: {head, from_list(tail)}

def to_list({}), do: []
def to_list({head, tail}), do: [head | to_list(tail)]
```

jiegillet:
Hi @ThiagosLima, since we have not received a PR from you since June, we would like to re-open this issue to other contributors. Of course if you are still working on it, feel free to let us know.

angelikatyborska:
[...]

PR #231, opened 2021-11-14
Add flatten-array analysis

angelikatyborska:
Closes https://github.com/exercism/elixir-analyzer/issues/60

I couldn't figure out any solutions that would use Stream only or list comprehensions only.

Website copy PR: https://github.com/exercism/website-copy/pull/2131

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -107,6 +107,9 @@ config :elixir_analyzer,
     "accumulate" => %{
       analyzer_module: ElixirAnalyzer.TestSuite.Accumulate
     },
+    "flatten-array" => %{
+      analyzer_module: ElixirAnalyzer.TestSuite.FlattenArray
+    },
     "leap" => %{
       analyzer_module: ElixirAnalyzer.TestSuite.Leap
     },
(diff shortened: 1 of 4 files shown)


jiegillet:
The implementation looks fine, but I looked at the problem description and nothing indicates that the problem should be solved with recursion only:
> Take a nested list and return a single flattened list with all values except nil/null.
> The challenge is to write a function that accepts an arbitrarily-deep nested list-like structure and returns a flattened structure without any nil/null values.

So before merging this, we should at least add an appendix to the instructions. I should have caught that on the opened issue, sorry. 
That being said, I'm not 100% convinced that we should prohibit using `Enum` et al. Maybe we can simply assert that `flatten` calls `flatten`, and call that recursion. For the record, here is my solution:

```elixir
  def flatten(element) when not(is_list(element)), do: [element]
  def flatten(list) do
    list
    |> Enum.map(&flatten/1)
    |> Enum.concat()
    |> Enum.reject(&is_nil/1)
  end
```

angelikatyborska:
> I looked at the problem description and nothing indicates that the problem should be solved with recursion only

Right... I'm not entirely sure why I opened the issue in the first place.

> Maybe we can simply assert that flatten calls flatten, and call that recursion.

I don't think we have a good way of asserting a function is recursive. This simple check wouldn't work if you have a private helper that's recursive.

I'm just going to let this exercise be as it is and close both the PR and issue.

PR #234, opened 2021-11-19
Fix high-score default argument check

angelikatyborska:


michallepicki (approved):

jiegillet (approved):
Thank you @michallepicki for bringing the issue up anf thank you @angelikatyborska for coding it up!

PR #235, opened 2021-11-20
Add common check for not using deprecated `:random` module

angelikatyborska:
Closes https://github.com/exercism/elixir/issues/1011

Website copy PR: https://github.com/exercism/website-copy/pull/2138

The PR contains a fix to ensure that all Elixir deprecations will be reported as compilation warnings.


lib/elixir_analyzer/test_suite/captains_log.ex
```diff
@@ -31,6 +31,7 @@ defmodule ElixirAnalyzer.TestSuite.CaptainsLog do
     calling_fn module: CaptainsLog, name: :random_stardate
     called_fn module: :rand, name: :uniform
     comment Constants.captains_log_use_rand_uniform()
+    suppress_if Constants.solution_deprecated_random_module()
```
angelikatyborska, line 34: @jiegillet I'm trying to suppress an exercise-specific check if a common check fails, but it doesn't work so far. Is it supposed to work?

jiegillet: Try this

```suggestion
    suppress_if Constants.solution_deprecated_random_module(), :fail
```

jiegillet (approved):
Sorry for the slow review, fiends and family weekend camping trip :)

Looks good to me (with the fix for `suppress_if`), but actually I have a suggestion.

Why don't we generalize that check to more deprecated modules?
Behaviour, Dict, GenEvent, HashDict ,HashSet, Set ,Supervisor.Spec are listed in the Elixir docs. Although I supposed those produce a warning already.
Then we could do [more Erlang modules](https://www.erlang.org/doc/general_info/deprecations.html#otp-24).

lib/elixir_analyzer/test_suite/captains_log.ex
angelikatyborska: Ah! Perfect.

For some reason I thought it needs to be a tuple, so I was attempting:
```elixir
suppress_if {Constants.solution_deprecated_random_module(), :fail}
```

which didn't work but also didn't give any hints about why

angelikatyborska:
> Why don't we generalize that check to more deprecated modules?

That's a good idea!

>  Behaviour, Dict, GenEvent, HashDict ,HashSet, Set ,Supervisor.Spec are listed in the Elixir docs. Although I supposed those produce a warning already.

I was convinced that would be the case because using e.g. `HashDict` produces a compilation warning... but the analyzer doesn't report it 😱 I need to figure out why

lib/elixir_analyzer/test_suite/captains_log.ex
jiegillet: Ah, maybe we should add an error message then, it wouldn't be too hard to do.

jiegillet:
Oh right, that's super weird. I thought maybe it's only at runtime, but nope.
```elixir
iex(66)> defmodule Old do        
...(66)> def foo, do: HashSet.new
...(66)> end
{:module, Old,
 <<70, 79, 82, 49, 0, 0, 4, 252, 66, 69, 65, 77, 65, 116, 85, 56,
   0, 0, 0, 152, 0, 0, 0, 16, 10, 69, 108, 105, 120, 105, 114, 46,
   79, 108, 100, 8, 95, 95, 105, 110, 102, 111, 95, 95, 10, 97,
   116, ...>>, {:foo, 0}}
iex(67)> Old.foo
#HashSet<[]>
iex(69)> HashSet.new
warning: HashSet.new/0 is deprecated. Use the MapSet module instead
  iex:69

#HashSet<[]>
```

angelikatyborska:
So far I noticed that those two calls give the warning:

```elixir
~S'''
fn -> HashDict.new() end
'''
|> Code.compile_string()


~S'''
defmodule MyModule do
  fn -> HashDict.new() end
end
'''
|> Code.compile_string()
```

But this last one, which would be the majority of cases for us, doesn't produce any warnings:

```
~S'''
defmodule MyModule do
  def foo(), do: HashDict.new()
end
'''
|> Code.compile_string()
```

jiegillet:
Well, an analyzer for this is really useful then. We could have one generic message "the following modules are deprecated: ...".

angelikatyborska:
I reached out for help on Elixir forum: https://elixirforum.com/t/why-does-code-compile-string-report-deprecation-warnings-only-sometimes/44012

If it doesn't yield anything useful soon, I'll start writing manual checks for all the modules we know are deprecated.

lib/elixir_analyzer/exercise_test/common_checks/compiler_warnings.ex
```diff
@@ -12,8 +12,10 @@ defmodule ElixirAnalyzer.ExerciseTest.CommonChecks.CompilerWarnings do
     warnings =
       capture_io(:stderr, fn ->
         try do
-          Code.compile_quoted(code_ast, Path.basename(code_path))
-          |> Enum.each(fn {module, _binary} ->
+          :elixir_compiler.quoted(code_ast, code_path, fn _, _ -> :ok end)
+          |> Enum.each(fn {module, map, binary} ->
+            Module.ParallelChecker.verify([{map, binary}], [])
+
```
angelikatyborska, line 18: With the hints from https://elixirforum.com/t/why-does-code-compile-string-report-deprecation-warnings-only-sometimes/44012/2 I came up with this solution. 

I took the compilations step from `compile_quoted`:
```
  def compile_quoted(quoted, file \\ "nofile") when is_binary(file) do
    loaded = :elixir_compiler.quoted(quoted, file, fn _, _ -> :ok end)
    Enum.map(loaded, fn {module, _map, binary} -> {module, binary} end)
  end
```

and then the verification step from `compile_file`
```elixir
  def compile_file(file, relative_to \\ nil) when is_binary(file) do
    loaded = :elixir_compiler.file(find_file(file, relative_to), fn _, _ -> :ok end)
    verify_loaded(loaded)
  end
```
```elixir
  defp verify_loaded(loaded) do
    maps_binaries = Enum.map(loaded, fn {_module, map, binary} -> {map, binary} end)
    Module.ParallelChecker.verify(maps_binaries, [])
    Enum.map(loaded, fn {module, _map, binary} -> {module, binary} end)
  end
```

I'm not sure if it's a great idea 🤔 the whole module `Module.ParallelChecker` is supposed to be private...

jiegillet: I don’t think it’s a good idea to rely on private modules… I found this thread the other day where José Valim was urging users not to do that, as it makes changing internal logic a lot more complex if people rely on it (I’m on my phone right now, I can find the link later)

instead, why don’t we compile the file directly? The part should be in the source. 
For testing, we could simply write a temp file at the same time as the source struct is being built.

jiegillet: Mmh maybe for testing it’s not that simple actually… maybe we can do that particular test from files only then?

jiegillet: Ah, it looks like I don't need to find the link I was talking about since José answered the same thing to your thread :)

angelikatyborska: Well, to me "private" code is more like social contract, and my interpretation of that contract is that I shouldn't use them, but if I really really have to, I have no right to complain when something unexpected happens 🙂. I opened a bug report for Elixir about this, but we don't need to wait for a fix because using `Code.compile_file` was trivial.

angelikatyborska:
To summarize:

> Why don't we generalize that check to more deprecated modules?
> Behaviour, Dict, GenEvent, HashDict ,HashSet, Set ,Supervisor.Spec are listed in the Elixir docs. Although I supposed those produce a warning already.

The PR contains a fix to ensure that all Elixir deprecations will be reported as compilation warnings.

> Then we could do more Erlang modules.

I don't feel the need to add more special checks for Erlang functions because all of the deprecated ones are unlikely to be needed on Exercism (except for `:random`). They are all about FTP, HTTP, SSH...

lib/elixir_analyzer/exercise_test/common_checks/deprecated_random_module.ex
```diff
@@ -0,0 +1,17 @@
+# credo:disable-for-this-file Credo.Check.Refactor.UnlessWithElse
```
jiegillet, line 1:
```suggestion
```

jiegillet (approved):
Nice. I'm glad the deprecated warning thing ended up being a trivial fix. 

I agree with you on the Erlang deprecated modules, it's unlikely that other will come up.

Issue #239, opened 2021-12-02
Analyzer is detecting a provided method as a method that should be private


jbcden:
In the remote-control-car exercise, there are 2 `new` methods provided (`new/0` and `new/1`). Both are needed to make the test suite pass (though you could replace `new/0` by adding a default value to `new/1`). When you submit the solution, the analyzer suggests making `new/0` private, even though it is called in the public interface.

While the analyzer is not necessarily wrong about `new/0` not being strictly necessary, I think this lint might be confusing to newer programmers and was wondering if we can find a way to prevent this lint or restructure the provided code to not have the `new/0`. Please let me know if anything is unclear or if I am misunderstanding the purpose of the analyzer suggestions.

For a reference to the provided code for this problem: https://github.com/exercism/elixir/blob/main/exercises/concept/remote-control-car/lib/remote_control_car.ex#L4-L6.

jiegillet:
Ah, good catch, thank you for reporting. 

I know what this happens. This check works by comparing the functions defined in the solution to the ones [in the exemplar implementation](https://github.com/exercism/elixir/blob/d00ca743340fcc328c70ee351274f91b57fd4c8d/exercises/concept/remote-control-car/.meta/exemplar.ex#L9-L11). Of course, in there it is done there by using a default argument, which is the idiomatic solution and produces a public `new/1` only. 

We claim that the functions that can be public are the ones used in the test suite, and `new/0` is definitely used. We could define a `new/0` in the exemplar but we would make it less idiomatic, so that's no good.

In the hints, we do mention "Consider a default-argument for the function", so I think we probably want to push students in that direction. We can do a custom check for the use of a default argument in `new/1` that would disable the "private functions" comment check. @angelikatyborska what do you think?

I could also tweak the common check so that when it sees a function `new(name \\ "default")` in the exemplar it produces both `new/0` and `new/1`, but I think then we would miss out on opportunities of adding custom checks and pushing students in the right direction, like in this case.

jbcden:
@jiegillet Thanks for looking into this! Good to know how this works :smiley:

angelikatyborska:
> We can do a custom check for the use of a default argument in new/1 that would disable the "private functions" comment check. We have one like that in two-fer already. @angelikatyborska what do you think?

Yes, that's a good idea!

> I could also tweak the common check so that when it sees a function new(name \\ "default") in the exemplar it produces both new/0 and new/1,

That is also a good idea.

[...]

PR #240, opened 2021-12-02
Tweak private function common check to consider default arguments

jiegillet:
Fixes one half of #239. The other half (adding a check for suggestion a default argument) is not covered here.

Now a function with `x` set arguments and `y` default arguments counts as `y+1` functions with arities ranging from `x` to `x+y`.

### DIFF
--- a/test/elixir_analyzer/exercise_test/common_checks/private_helper_functions_test.exs
+++ b/test/elixir_analyzer/exercise_test/common_checks/private_helper_functions_test.exs
@@ -13,6 +13,10 @@ defmodule ElixirAnalyzer.ExerciseTest.CommonChecks.PrivateHelperFunctionsTest do
                 |> File.read!()
                 |> Code.string_to_quoted()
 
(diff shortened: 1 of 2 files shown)


test/elixir_analyzer/exercise_test/common_checks/private_helper_functions_test.exs
```diff
@@ -486,4 +490,96 @@ defmodule ElixirAnalyzer.ExerciseTest.CommonChecks.PrivateHelperFunctionsTest do
       assert PrivateHelperFunctions.run(code, @sqrt_example) == [{:fail, comment}]
     end
   end
+
+  describe "concept exercise with remote-control-car" do
+    # This exemplar has a default argument: def new(nickname \\ "none")
```
angelikatyborska, line 495: Could we copy-paste the exemplar to `@remote_control_car_exemplar` instead of loading it from a file? If that happened around this line of code instead of at the very beginning of the file, we might not need this comment at all. The test will also continue working in case we ever change the exercise.

angelikatyborska (approved):

PR #241, opened 2021-12-02
Add analyzer check for `remote-control-car`

### LINKED ISSUE #239: Analyzer is detecting a provided method as a method that should be private

jiegillet:
Closes #239.
[Comment PR here](https://github.com/exercism/website-copy/pull/2140).

The CI tests cannot pass until #240 is merged because the problem described in #239.

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -84,6 +84,9 @@ config :elixir_analyzer,
     "pacman-rules" => %{
       analyzer_module: ElixirAnalyzer.TestSuite.PacmanRules
     },
+    "remote-control-car" => %{
+      analyzer_module: ElixirAnalyzer.TestSuite.RemoteControlCar
+    },
     "rpg-character-sheet" => %{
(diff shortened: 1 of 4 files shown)


angelikatyborska (approved):
pretty straight-forward!

Issue #84, opened 2021-07-07
 Write analysis for concept exercise `dna-encoding`


angelikatyborska:
Desired checks:
https://github.com/exercism/elixir/blob/9774308bb6b898df6ecbde64c5c6e48c651870ae/exercises/concept/dna-encoding/.meta/design.md#L29-L32

jiegillet:
This can de done using existing tools.

julngomz:
I can work on this issue.

jiegillet:
Hello!
We haven’t seen any PR for this, so feel free to work on it.

ryanzidago:
> Hello! We haven’t seen any PR for this, so feel free to work on it.

I've opened up a PR https://github.com/exercism/elixir-analyzer/pull/242
Let me know what you think :pray:

PR #242, opened 2021-12-03
Add analyzer module for DNA encoding exercise

ryanzidago:
Issue: https://github.com/exercism/elixir-analyzer/issues/84.

# Description:
I added checks that will prompt users to use `recursion` instead of the following modules/functions:
- `Enum.reduce`
- `List.foldr`
- `for`

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -39,6 +39,9 @@ config :elixir_analyzer,
     "chessboard" => %{
       analyzer_module: ElixirAnalyzer.TestSuite.Chessboard
     },
+    "dna-encoding" => %{
+      analyzer_module: ElixirAnalyzer.TestSuite.DNA
+    },
     "file-sniffer" => %{
       analyzer_module: ElixirAnalyzer.TestSuite.FileSniffer
     },
--- a/config/config.exs
+++ b/config/config.exs
@@ -40,7 +40,7 @@ config :elixir_analyzer,
       analyzer_module: ElixirAnalyzer.TestSuite.Chessboard
     },
     "dna-encoding" => %{
-      analyzer_module: ElixirAnalyzer.TestSuite.DNA
+      analyzer_module: ElixirAnalyzer.TestSuite.DNAEncoding
     },
     "file-sniffer" => %{
       analyzer_module: ElixirAnalyzer.TestSuite.FileSniffer
--- a/lib/elixir_analyzer/test_suite/dna_encoding.ex
+++ b/lib/elixir_analyzer/test_suite/dna_encoding.ex
@@ -6,19 +6,25 @@ defmodule ElixirAnalyzer.TestSuite.DNAEncoding do
   use ElixirAnalyzer.ExerciseTest
   alias ElixirAnalyzer.Constants
 
-  assert_no_call "does not call Enum.reduce" do
+  assert_no_call "does not call any Enum functions" do
     type :essential
-    called_fn module: Enum, name: :reduce
+    called_fn module: Enum, name: :_
     comment Constants.dna_encoding_use_recursion()
   end
 
-  assert_no_call "does not call List.foldr" do
+  assert_no_call "does not call any Stream functions" do
     type :essential
-    called_fn module: List, name: :foldr
+    called_fn module: Stream, name: :_
     comment Constants.dna_encoding_use_recursion()
   end
 
-  assert_no_call "does not call the for/1 function" do
+  assert_no_call "does not call any List functions" do
+    type :essential
+    called_fn module: List, name: :_
+    comment Constants.dna_encoding_use_recursion()
+  end
+
+  assert_no_call "does not use list comprehensions" do
     type :essential
     called_fn name: :for
     comment Constants.dna_encoding_use_recursion()
--- a/lib/elixir_analyzer/test_suite/dna_encoding.ex
+++ b/lib/elixir_analyzer/test_suite/dna_encoding.ex
@@ -1,0 +1,26 @@
+defmodule ElixirAnalyzer.TestSuite.DNA do
+  @moduledoc """
+  This is an exercise analyzer extension module for the DNA Encoding concept exercise
+  """
+
+  use ElixirAnalyzer.ExerciseTest
+  alias ElixirAnalyzer.Constants
+
+  assert_no_call "does not call Enum.reduce/{2,3}" do
+    type :essential
+    called_fn module: Enum, name: :reduce
+    comment Constants.dna_encoding_use_recursion()
+  end
+
+  assert_no_call "does not call List.foldr" do
+    type :essential
+    called_fn module: List, name: :foldr
+    comment Constants.dna_encoding_use_recursion()
+  end
+
+  assert_no_call "does not call the for/1 function" do
+    type :essential
+    called_fn name: :for
+    comment Constants.dna_encoding_use_recursion()
+  end
+end
--- a/lib/elixir_analyzer/test_suite/dna_encoding.ex
+++ b/lib/elixir_analyzer/test_suite/dna_encoding.ex
@@ -6,7 +6,7 @@ defmodule ElixirAnalyzer.TestSuite.DNA do
   use ElixirAnalyzer.ExerciseTest
   alias ElixirAnalyzer.Constants
 
-  assert_no_call "does not call Enum.reduce/{2,3}" do
+  assert_no_call "does not call Enum.reduce" do
     type :essential
     called_fn module: Enum, name: :reduce
     comment Constants.dna_encoding_use_recursion()
--- a/lib/elixir_analyzer/test_suite/dna_encoding.ex
+++ b/lib/elixir_analyzer/test_suite/dna_encoding.ex
@@ -1,4 +1,4 @@
-defmodule ElixirAnalyzer.TestSuite.DNA do
+defmodule ElixirAnalyzer.TestSuite.DNAEncoding do
   @moduledoc """
   This is an exercise analyzer extension module for the DNA Encoding concept exercise
   """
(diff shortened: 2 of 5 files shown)


config/config.exs
```diff
@@ -39,6 +39,9 @@ config :elixir_analyzer,
     "chessboard" => %{
       analyzer_module: ElixirAnalyzer.TestSuite.Chessboard
     },
+    "dna-encoding" => %{
+      analyzer_module: ElixirAnalyzer.TestSuite.DNA
```
jiegillet, line 43: Could you call the module `DnaEncoding`? Or `DNAEncoding`? The capitalization is a bit awkward but we need the `Encoding` for consistency.

test/elixir_analyzer/test_suite/dna_encoding_test.exs
```diff
@@ -0,0 +1,180 @@
...
+      def decode(dna) do
+        do_decode(dna, [])
+      end
+
+      defp do_decode(<<>>, acc), do: acc |> reverse()
+
+      defp do_decode(<<n::4, rest::bitstring>>, acc),
+        do: do_decode(rest, [decode_nucleotide(n) | acc])
+
+      defp reverse(l), do: do_reverse(l, [])
```
jiegillet, line 53: @angelikatyborska do you think we should ban `Enum.reverse` as well? I'm leaning towards yes.

test/elixir_analyzer/test_suite/dna_encoding_test.exs
```diff
@@ -0,0 +1,180 @@
...
+        def decode_nucleotide(0b0000), do: ?\s
+        def decode_nucleotide(0b0001), do: ?A
+        def decode_nucleotide(0b0010), do: ?C
+        def decode_nucleotide(0b0100), do: ?G
+        def decode_nucleotide(0b1000), do: ?T
+
+        def encode(dna) do
+          dna
+          |> Enum.map(&encode_nucleotide/1)
+          |> Enum.reduce("", fn enc, bs -> <<bs::bitstring, enc::4>> end)
```
jiegillet, line 146: If you use `Enum.reduce` here, we can't know if the comment is triggered by it or by `for`.

lib/elixir_analyzer/test_suite/dna_encoding.ex
```diff
@@ -0,0 +1,26 @@
...
+
+  assert_no_call "does not call List.foldr" do
+    type :essential
+    called_fn module: List, name: :foldr
+    comment Constants.dna_encoding_use_recursion()
+  end
+
+  assert_no_call "does not call the for/1 function" do
+    type :essential
+    called_fn name: :for
```
jiegillet, line 23: Here, students would still be able to use it in the form of `Kernel.SpecialForms.for`. I know it's a bit of  stretch that they would do that, but there is a simple way to add it. First, change to

```suggestion
    called_fn module: Kernel.SpecialForm, name: :for
```

Then we have to make sure that the `Kernel.SpecialForm` functions are imported, you can do that here:
https://github.com/exercism/elixir-analyzer/blob/71130beddcab0da34896ea41f24ba65264437eae/lib/elixir_analyzer/exercise_test/assert_call/compiler.ex#L57
(do to `[:Kernel, :SpecialForm]` what is done to `[:Kernel]`).

Hopefully that works, let me know if it doesn't :D

jiegillet (commented):
Thank you that's awesome!
I left a few comments, one of them is a bit tricky, so ask me if you need more help.

jiegillet:
I almost forgot, you will also need to create a PW on website-copy ([like this one](https://github.com/exercism/website-copy/pull/2131)) to add the analyzer comment.

config/config.exs
ryanzidago: I was not sure about this, as the module of the exercise is named `DNA` as you can see [here](https://github.com/exercism/elixir/blob/d00ca743340fcc328c70ee351274f91b57fd4c8d/exercises/concept/dna-encoding/lib/dna.ex#L1) and I thought that the anaylzer always refered to an existing module :thinking: (but the project/directory/exercise is named `dna-encoding`).

Should I leave `DNA` or rename it to `DNAEncoding`?

lib/elixir_analyzer/test_suite/dna_encoding.ex
ryanzidago: I've changed the `modules_in_scope` value in the `assert/3` function, in compiler.ex to:
```elixir
modules_in_scope: %{
        [:Kernel] => Kernel.module_info(:exports),
        [:Kernel, :SpecialForms] => Kernel.SpecialForms.module_info(:exports)
      }
```

And the test suite to:
```elixir
assert_no_call "does not call the for/1 function" do
    type :essential
    called_fn module: Kernel.SpecialForms, name: :for
    comment Constants.dna_encoding_use_recursion()
  end
```

However, I still get two tests failing as `"elixir.dna-encoding.use_recursion"` isn't returned anymore :thinking:

ryanzidago:
> I almost forgot, you will also need to create a PW on website-copy ([like this one](https://github.com/exercism/website-copy/pull/2131)) to add the analyzer comment.

Roger that. I just realized that the standard message for informing students to use recursion is the following:
> The purpose of this exercise is to teach recursion. Solve it without using list comprehensions or any of the functions from the modules `Enum`, `List`, or `Stream`.

Then, I should probably extend this PR to also include the Stream module and other functions rather `reduce`, `for` and `foldr`? :thinking:

test/elixir_analyzer/test_suite/dna_encoding_test.exs
angelikatyborska: > @angelikatyborska do you think we should ban Enum.reverse as well? I'm leaning towards yes.

> Then, I should probably extend this PR to also include the Stream module and other functions rather reduce, for and foldr? 🤔

Since this is an exercise for learning tail-call recursion, I think we should ban all Enum , List, and Stream functions, as well as list comprehensions. Maybe it's not perfect to ban innocent List functions like `first` or `last`, but they're not really needed and using them instead of pattern matching is not idiomatic, so I wouldn't care 🤷

lib/elixir_analyzer/test_suite/dna_encoding.ex
jiegillet: I looked into this more, and it seems I was wrong. `for` is special and cannot in fact be called with the fill prefix.

```elixir
iex(66)> for a <- [1,2,3], do: a
[1, 2, 3]
iex(67)> Kernel.SpecialForms.for a <- [1,2,3], do: a
** (CompileError) iex:67: undefined function <-/2
    (stdlib 3.15) lists.erl:1358: :lists.mapfoldl/3
```

The technical reason why this didn't work here is that `Kernel.SpecialForms` exposes `for` with arity 1, which didn't match the example (each `<-` and `do` add an argument). 

LEt's keep it the way you had it, it seems I have wasted your time on this one, I apologize.

test/elixir_analyzer/test_suite/dna_encoding_test.exs
jiegillet: Sounds reasonable. @ryanzidago can you change the checks to do that?

config/config.exs
jiegillet: The exercise itself can break convention, but here we are referencing many exercises, so I think I would prefer having the full name.

test/elixir_analyzer/test_suite/dna_encoding_test.exs
```diff
@@ -0,0 +1,190 @@
...
+          do: do_decode(rest, [decode_nucleotide(n) | acc])
+
+        defp reverse(l), do: do_reverse(l, [])
+        defp do_reverse([], acc), do: acc
+        defp do_reverse([h | t], acc), do: do_reverse(t, [h | acc])
+      end
+    ]
+  end
+
+  test_exercise_analysis "detects the usage of the List module",
```
jiegillet, line 97: Can you add one test with the `Stream` module?

ryanzidago: Done! :white_check_mark:

test/elixir_analyzer/test_suite/dna_encoding_test.exs
```diff
@@ -0,0 +1,226 @@
+defmodule ElixirAnalyzer.ExerciseTest.DNAEncodingTest do
+  use ElixirAnalyzer.ExerciseTestCase,
+    exercise_test_module: ElixirAnalyzer.TestSuite.DNAEncoding
+
+  test_exercise_analysis "example solution",
+    comments: [ElixirAnalyzer.Constants.solution_same_as_exemplar()] do
+    defmodule DNA do
+      @moduledoc """
```
jiegillet, line 8: @angelikatyborska note for later, this moduledoc will make it impossible for anyone to get the celebratory comment. Once the representer goes live, we should consider comparing solutions through it. In the mean time, maybe we can remove the moduledoc in the exemplar?

jiegillet (approved):
Great, thank you very much for your contribution!

We very recently switched this repo to use Elixir 1.13, and one of the changes made the tests break, so I took the liberty of committing a patch to your branch. 

The comment on website-copy was in the wrong folder, so the external tests still fail. I [submitted a PR](https://github.com/exercism/website-copy/pull/2143), as soon as it goes in, we can merge this one!

test/elixir_analyzer/test_suite/dna_encoding_test.exs
angelikatyborska: Of course! I'm surprised it's even there. I created a PR https://github.com/exercism/elixir/pull/1020

Issue #229, opened 2021-11-10
Dependabot for Elixir deps for all tooling


angelikatyborska:
Should we configure all tooling repos to automatically update Elixir deps with dependabot?

- Analyzer uses: jason, credo, dialyzer
- Test Runner uses: jason
- Representer uses: nothing

neenjaw:
depends what update means? patch versions? sure. Minor/major versions? no, I think that should be done by a human

jiegillet:
I don't have much experience with long term maintenance of dependencies, but in principle I don't have an issue with asking dependabot to update things. If all tests pass, that should be fine, we can always fix broken things later. We can also treat dependabot as a notifier of new versions and not feel pressured to update...

angelikatyborska:
In general, if you intend to maintain a project long term, I would say some sort of automated dependency update is necessary. The reason is that if you do maintain a project, at some point you will need to do a change to the project that requires a new external dependency or a new feature from the current one. If you didn't do regular dep updates, now you have to do all of them at once. This increases the risk and time sink of that one change.

I think our tools will be maintained long term, and we need fresh dependencies because we will do Elixir lang updates in all of them regularly.

If you have a good test coverage, automated dep updates should be safe. If you don't, you need to test each one manually. Here is where I am not so sure about our situation 🤔 

The situation also hugely varies between Elixir deps and JS deps. Thankfully Elixir world is much more sane, and the version level changes actually mean something. In JS world, there are plenty of deps where major version updates aren't actually that, and plenty of patch version updates break your whole project 🤦 

I think my vote would be to add automated updates to our projects for patch and minor updates, but not for major updates. But also there's so few of them, that I can also see this whole thing not being worth it.

jiegillet:
How do we evaluate test coverage?

angelikatyborska:
In the past I have used https://coveralls.io/ + https://github.com/parroty/excoveralls, apparently you can also pass a `--cover` option to `mix test`:

```
$ mix test --include external --cover
Cover compiling modules ...
Excluding tags: [:pending]
Including tags: [:external]

.....................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................# 2021-11-24T17:36:37.864Z [] [warn] Description: 'Authenticity is not established by certificate path validation'
     Reason: 'Option {verify, verify_peer} and cacertfile/cacerts is missing'

[...]

PR #246, opened 2021-12-18
100% test coverage

jiegillet:
In #229, there was a discussion about automated dependency updates being safer with better test coverage, so I took it as a learning opportunity and chipped away at it and reached full coverage.

The type of changes changes I made were:
- Adding tests
- Removing dead code
- Refactoring
- Fixing bugs discovered by the extra tests
- Duplicating some tests for `feature`, `assert_call` and `check_source` because for some reason the modules with `use ElixirAnalyzer.ExerciseTest` are not always taken into account if they are not defined in a `*_test.exs` file even if they are actually ran. This is probably due to the heavy macro use. These are a bit unfortunate.

I will add some comments on some parts where I have questions a bit later.

I used `excoveralls` to find the details of what was or wasn't covered, but I didn't commit that change not to add a dependency. I could do it though, and we could use https://coveralls.io/  to get a badge. I'd be happy to add that (in a different PR, I guess).

### DIFF
--- a/lib/elixir_analyzer.ex
+++ b/lib/elixir_analyzer.ex
@@ -141,14 +141,14 @@ defmodule ElixirAnalyzer do
       }
     rescue
       e in File.Error ->
-        Logger.warning("Unable to decode 'config.json'", error_message: e.message)
+        Logger.warning("Unable to read config file #{e.path}", error_message: e.reason)
 
         submission
         |> Submission.halt()
         |> Submission.set_halt_reason("Analysis skipped, not able to read solution config.")
 
       e in Jason.DecodeError ->
-        Logger.warning("Unable to decode 'config.json'", error_message: e.message)
+        Logger.warning("Unable to decode 'config.json'", data: e.data)
 
         submission
         |> Submission.halt()
--- a/lib/elixir_analyzer.ex
+++ b/lib/elixir_analyzer.ex
@@ -29,17 +29,13 @@ defmodule ElixirAnalyzer do
 
   * `exercise` is which exercise is submitted to determine proper analysis
 
-  * `path` is the path (ending with a '/') to the submitted solution
+  * `input_path` is the path to the submitted solution
+
+  * `output_path` is the path to the output folder
 
   * `opts` is a Keyword List of options, see **options**
 
   ## Options
-
-  * `:exercise` - name of the exercise, defaults to the `exercise` parameter
-
-  * `:path` - path to the submitted solution, defaults to the `path` parameter
-
-  * `:output_path` - path to write file output, defaults to the `path` parameter
 
   * `:output_file`, - specifies the name of the output_file, defaults to
     `@output_file` (`analysis.json`)
@@ -52,8 +48,6 @@ defmodule ElixirAnalyzer do
 
   * `:puts_summary` - boolean flag if an analysis should print the summary of the
     analysis to stdio, defaults to `true`
-
-  Any arbitrary keyword-value pair can be passed to `analyze_exercise/3` and these options may be used the other consuming code.
   """
   @spec analyze_exercise(String.t(), String.t(), String.t(), keyword()) :: Submission.t()
   def analyze_exercise(exercise, input_path, output_path, opts \\ []) do
@@ -256,7 +250,7 @@ defmodule ElixirAnalyzer do
 
     submission =
       submission
-      |> submission.analysis_module.analyze(submission.source)
+      |> submission.analysis_module.analyze()
       |> Submission.set_analyzed(true)
 
     Logger.info("Analyzing code complete")
--- a/lib/elixir_analyzer/cli.ex
+++ b/lib/elixir_analyzer/cli.ex
@@ -6,22 +6,19 @@ defmodule ElixirAnalyzer.CLI do
   @usage """
   Usage:
 
-    $ elixir_analyzer <exercise-name> <input path> <output path>  [options]
+    $ elixir_analyzer <exercise-name> <input path> <output path> [options]
 
   You may also pass the following options:
-    --skip-analysis                       flag skips running the static analysis
-    --output-file <filename>
-
-  You may also test only individual files :
-    (assuming analyzer tests are compiled for the named module)
-
-    $ exercism_analyzer --analyze-file <full-path-to-.ex>:<module-name>
+    --help                          see this message
+    --output-file <filename>        output file name (default: analysis.json)
+    --no-write-results              doesn't write to JSON file
+    --no-puts-summary               doesn't print summary to stdio
   """
 
   @options [
-    {{:skip_analyze, :boolean}, false},
     {{:output_file, :string}, "analysis.json"},
-    {{:analyze_file, :string}, nil},
+    {{:write_results, :boolean}, true},
+    {{:puts_summary, :boolean}, true},
     {{:help, :boolean}, false}
   ]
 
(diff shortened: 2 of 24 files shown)


lib/elixir_analyzer.ex
```diff
@@ -29,18 +29,14 @@ defmodule ElixirAnalyzer do
...
 
   * `opts` is a Keyword List of options, see **options**
 
   ## Options
 
-  * `:exercise` - name of the exercise, defaults to the `exercise` parameter
-
-  * `:path` - path to the submitted solution, defaults to the `path` parameter
-
-  * `:output_path` - path to write file output, defaults to the `path` parameter
```
jiegillet, line 42: I did not understand these options here, all they do is overwrite the required arguments. Those options were passed from the CLI module. I got rid of them.

lib/elixir_analyzer/cli.ex
```diff
@@ -6,22 +6,19 @@ defmodule ElixirAnalyzer.CLI do
...
+    $ elixir_analyzer <exercise-name> <input path> <output path> [options]
 
   You may also pass the following options:
-    --skip-analysis                       flag skips running the static analysis
-    --output-file <filename>
-
-  You may also test only individual files :
-    (assuming analyzer tests are compiled for the named module)
-
-    $ exercism_analyzer --analyze-file <full-path-to-.ex>:<module-name>
```
jiegillet, line 18: This was an option in the CLI, but could not actually be passed to the analyzer, so I got rid of it

lib/elixir_analyzer/cli.ex
```diff
@@ -6,22 +6,19 @@ defmodule ElixirAnalyzer.CLI do
   @usage """
   Usage:
 
-    $ elixir_analyzer <exercise-name> <input path> <output path>  [options]
+    $ elixir_analyzer <exercise-name> <input path> <output path> [options]
 
   You may also pass the following options:
-    --skip-analysis                       flag skips running the static analysis
```
jiegillet, line 12: I don't understand the purpose of skipping analysis except for debugging, so I got rid of that option too. I think it still exists somewhere in `feature` though, although it's never used.

lib/elixir_analyzer/quote_util.ex
```diff
@@ -68,15 +68,6 @@ defmodule ElixirAnalyzer.QuoteUtil do
     end)
   end
 
-  @doc """
-  Performs a depth-first, pre-order traversal of quoted expressions.
-  With depth provided to a function
-  """
-  @spec prewalk(Macro.t(), (Macro.t(), non_neg_integer -> Macro.t())) :: Macro.t()
-  def prewalk(ast, fun) when is_function(fun, 2) do
```
jiegillet, line 76: I was a bit hesitant to get rid of this. It seems like it could be useful, but it's not actually used anywhere. I'm assuming at this point we won't need to add another complex analysis tool.

test/elixir_analyzer/exercise_test/check_source_test.exs
```diff
@@ -121,4 +121,89 @@ defmodule ElixirAnalyzer.ExerciseTest.CheckSourceTest do
       """
     ]
   end
+
+  test "full check_source definition in a test file for coverage" do
+    defmodule Coverage do
```
jiegillet, line 126: This is one of these cases where even though these things are being used somewhere else, somehow the coverage tool doesn't see it. It's a bit messy, but I suppose it doesn't hurt to have redundant tests.

test/elixir_analyzer_test.exs
```diff
@@ -150,6 +170,111 @@ defmodule ElixirAnalyzerTest do
     end
   end
 
+  describe "different failures" do
+    test "summary for a submission that did not run" do
```
jiegillet, line 174: This is a bit artificial, because I needed to find a test that gives a report with an incomplete analysis, but could not really find a branching in the code that would lead to this. From the point of view of `Summary.summary/2` it makes sense to have that formatting option though...

test/elixir_analyzer/exercise_test/feature_test.exs
```diff
@@ -382,4 +382,227 @@ defmodule ElixirAnalyzer.ExerciseTest.FeatureTest do
...
+    end
+  end
+
+  describe "other features" do
+    defmodule Coverage do
+      use ElixirAnalyzer.ExerciseTest
+
+      feature "skip" do
+        comment "skip"
+        status :skip
```
jiegillet, line 455: Do we really need to keep that `status` field? It might have been useful for debugging, but we don't actually use it anywhere in the code. I think it's commented out in `two-fer`.

test/elixir_analyzer/exercise_test/feature_test.exs
```diff
@@ -382,4 +382,227 @@ defmodule ElixirAnalyzer.ExerciseTest.FeatureTest do
...
+        depth 3
+
+        form do
+          :ok
+        end
+      end
+
+      feature "meta" do
+        comment "meta"
+        meta(keep_meta(true))
```
jiegillet, line 517: Same here, do we need this `meta` field? 
I can't imagine an application where we would want to keep the meta information. In fact I'm pretty sure that setting `keep_meta(true)` is equivalent to "never match this form".

lib/elixir_analyzer.ex
```diff
@@ -256,7 +250,7 @@ defmodule ElixirAnalyzer do
 
     submission =
       submission
-      |> submission.analysis_module.analyze(submission.source)
+      |> submission.analysis_module.analyze()
```
angelikatyborska, line 253: Nice getting rid of a duplicated argument 👍

lib/elixir_analyzer/cli.ex
angelikatyborska: > I think it still exists somewhere in feature though, although it's never used.

Do you mean the ability to skip a single check? Because I don't see anything related to skipping analysis as a whole anywhere else 🤔

lib/elixir_analyzer/exercise_test/common_checks/function_capture.ex
```diff
@@ -60,7 +60,7 @@ defmodule ElixirAnalyzer.ExerciseTest.CommonChecks.FunctionCapture do
         functions
       end
 
-    {node, %{capture_depth: depth - 1, functions: functions}}
+    {node, %{capture_depth: depth, functions: functions}}
```
angelikatyborska, line 63: How did you discover this bug? When I bring it back, no tests fail, so it wasn't by writing tests 🤔

lib/elixir_analyzer/exercise_test/feature.ex
```diff
@@ -53,10 +53,16 @@ defmodule ElixirAnalyzer.ExerciseTest.Feature do
     feature_data = %{feature_data | meta: Map.to_list(feature_data.meta)}
     feature_data = Map.to_list(feature_data)
 
+    unless Keyword.has_key?(feature_data, :comment) do
+      raise "Comment must be defined for each feature test"
+    end
```
angelikatyborska, line 58: Awesome 🎉  I didn't notice before that this one macro is missing the comment check

lib/elixir_analyzer/quote_util.ex
angelikatyborska: No mercy for dead code! Even if it looks potentially useful, we will be able to recreate it when we need it. Making the code base easier to understand by removing trash is the better option every time, IMO.

test/elixir_analyzer/exercise_test/feature_test.exs
angelikatyborska: My guess is that it would be safe to remove `status`. @neenjaw do you have any thoughts about this?

test/elixir_analyzer/exercise_test/feature_test.exs
angelikatyborska: I think we can delete it. Forcing meta to match would mean forcing line numbers to match, which doesn't seem useful to me for anything. Adding an empty new line anywhere would mess up the whole analysis? Nah.

angelikatyborska (approved):
Excellent job and very valuable changes 👏. Only one more fix to `README.md` is necessary before merging.

If you want to keep removing stuff (skip and metadata for example), can you open a new PR for that?

> I could do it though, and we could use [coveralls.io](https://coveralls.io/) to get a badge. I'd be happy to add that (in a different PR, I guess).

Totally, go for it!

test/elixir_analyzer/exercise_test/feature_test.exs
jiegillet: Ok, I'll get rid of it. In a different PR :)

lib/elixir_analyzer/exercise_test/common_checks/function_capture.ex
jiegillet: Sorry about that! I took me a while to remember.
I think it was from this test:

https://github.com/exercism/elixir-analyzer/blob/926d69d60f13ccee353520a09e9c882c53b0e4d2/test/elixir_analyzer/exercise_test/common_checks/function_capture_test.exs#L53-L58

At some point I was checking if `actual_function?` gave the right result, I was printing some values and I was expecting to see that test case pop out, but it never came, so I realized that `depth` was becoming negative.

I've added a failing test now. Let that be a valuable lesson of never fixing bugs without context.

jiegillet:
OK, I think I've addressed everything, please check my last 2 commits.
I'll wait for @neenjaw's input before removing `:status` but it would be in a different PR anyway.

Like always, thank you for your incredible thoroughness in your reviews, especially for this huge one!

Yay 100%!

PR #252, opened 2022-02-03
Remove protocol consolidation to fix warning in valid code

### LINKED ISSUE #251: `clock` produces protocol consolidation compiler warning
Even for the example implementation,
```elixir
  defimpl String.Chars, for: Clock do
    def to_string(%Clock{hour: hour, minute: minute}) do
      "#{format(hour)}:#{format(minute)}"
    end 

    defp format(number) do
      number |> Integer.to_string() |> String.pad_leading(2, "0")
    end 
  end 
```

A warning is produced 
```
warning: the String.Chars protocol has already been consolidated, an implementation for Clock has no effect. If you want to implement protocols after compilation or during tests, check the "Consolidation" section in the Protocol module documentation
  /mnt/exercism-iteration/lib/clock.ex:16: Clock (module)
```

There is no warning when compiling the file locally.

This could be due to `ElixirAnalyzer.ExerciseTest.CommonChecks.CompilerWarnings` somehow re-compiling the code several times and not purging the protocol consolidation.

jiegillet:
Closes #251.

My initial hypothesis (in the issue) was not quite right. [Protocol consolidation](https://hexdocs.pm/elixir/Protocol.html#module-consolidation) is a feature that links protocols to their implementation in order to speed up the protocol dispatch. This can only work when all implementations are known, typically at compile time.

In the case of `clock`, when we compile the file in `ElixirAnalyzer.ExerciseTest.CommonChecks.CompilerWarnings`, an extra implementation is defined and the compiler refuses to include it because the "protocol has already been consolidated" at compile time. 

The docs only show how to turn it off for tests, so I brought my hammer for this fix and I turned off the consolidation altogether. The tradeoff is supposedly a performance cost, but I didn't notice any difference. I'm of course open to other ideas.

### DIFF
--- a/test/elixir_analyzer/exercise_test/common_checks/compiler_warnings_test.exs
+++ b/test/elixir_analyzer/exercise_test/common_checks/compiler_warnings_test.exs
@@ -1,0 +1,9 @@
+defmodule ElixirAnalyzer.ExerciseTest.CommonChecks.CompilerWarningsTest do
+  use ExUnit.Case
+  alias ElixirAnalyzer.ExerciseTest.CommonChecks.CompilerWarnings
+
+  test "Implementing a protocol doesn't trigger a compiler warning" do
+    filepath = "test_data/clock/lib/clock.ex"
+    assert CompilerWarnings.run(filepath, nil) == []
+  end
+end
--- a/mix.exs
+++ b/mix.exs
@@ -8,6 +8,8 @@ defmodule ElixirAnalyzer.MixProject do
       elixir: "~> 1.13",
       elixirc_paths: elixirc_paths(Mix.env()),
       start_permanent: Mix.env() == :prod,
+      # Turn off protocol consolidation to avoid warning in analyzed code
+      consolidate_protocols: false,
       deps: deps(),
       escript: escript(),
       preferred_cli_env: [
--- a/test_data/clock/lib/clock.ex
+++ b/test_data/clock/lib/clock.ex
@@ -1,0 +1,60 @@
+defmodule Clock do
+  defstruct hour: 0, minute: 0
+  @type t :: %Clock{hour: integer, minute: integer}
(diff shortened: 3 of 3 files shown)


test/elixir_analyzer/exercise_test/common_checks/compiler_warnings_test.exs
```diff
@@ -0,0 +1,9 @@
+defmodule ElixirAnalyzer.ExerciseTest.CommonChecks.CompilerWarningsTest do
+  use ExUnit.Case
+  alias ElixirAnalyzer.ExerciseTest.CommonChecks.CompilerWarnings
+
+  test "Implementing a protocol doesn't trigger a compiler warning" do
+    filepath = "test_data/clock/lib/clock.ex"
+    assert CompilerWarnings.run(filepath, nil) == []
+  end
```
angelikatyborska, line 8: How can I get this test to fail? Removing `consolidate_protocols: false,` form `mix.exs` doesn't work.

angelikatyborska:
Initially I was worried about this because I remembered something about protocol consolidation being crucial to one exercise, but thankfully that was in the test runner: https://github.com/exercism/elixir-test-runner/pull/51

test/elixir_analyzer/exercise_test/common_checks/compiler_warnings_test.exs
angelikatyborska: Nevermind, I was editing `elixir/mix.exs` 🤦‍♀️

angelikatyborska (approved):

test/elixir_analyzer/exercise_test/common_checks/compiler_warnings_test.exs
jiegillet: Hahaha, yeah, that won't work :)

PR #253, opened 2022-02-03
Only comment on explicit `:ok` in `rpg-character-sheet` and `newsletter`

### LINKED ISSUE #250: RPG character sheet falsely returns errors when importing IO
A solution that imports `IO` is technically correct, but it fails analysis because `welcome` doesn't end in `IO.puts`.

```elixir
  import IO

  def welcome() do
    puts("Welcome! Let's fill out your character sheet together.")
  end

  def ask_name() do
    gets("What is your character's name?\n") |> String.trim()
  end

  def ask_class() do
    gets("What is your character's class?\n") |> String.trim()
  end

  def ask_level() do
    gets("What is your character's level?\n") |> String.trim() |> String.to_integer()
  end

  def run() do
    welcome()
    name = ask_name()
    class = ask_class()
    level = ask_level()

    %{
      name: name,
      class: class,
      level: level
    }
    |> IO.inspect(label: "Your character")
  end
end
```

Ideally both imported and aliased usages of IO functions would be supported by the analyzer.

jiegillet:
Closes #250.

The issue only mentioned `rpg-character-sheet` but `newsletter` was using the same pattern so I changed it too.

This is not the first time those comments are misleading students, so I thought it would be best to only send a comment to solutions we 100% know are not what we want. It does mean we might be missing some cases (like the tests I had to remove), hopefully not too many. 

I think it's a fair tradeoff, but I'm happy to discuss that choice.

### DIFF
--- a/test/elixir_analyzer/test_suite/rpg_character_sheet_test.exs
+++ b/test/elixir_analyzer/test_suite/rpg_character_sheet_test.exs
@@ -145,6 +145,39 @@ defmodule ElixirAnalyzer.TestSuite.RpgCharacterSheetTest do
           |> IO.gets()
           |> String.trim()
         end
+      end,
+      defmodule RPG.CharacterSheet do
+        import IO
+
+        def welcome() do
+          puts("Welcome! Let's fill out your character sheet together.")
+        end
+
+        def ask_name() do
+          gets("What is your character's name?\n") |> String.trim()
+        end
+
+        def ask_class() do
+          gets("What is your character's class?\n") |> String.trim()
+        end
+
+        def ask_level() do
+          gets("What is your character's level?\n") |> String.trim() |> String.to_integer()
+        end
+
+        def run() do
+          welcome()
+          name = ask_name()
+          class = ask_class()
+          level = ask_level()
+
+          %{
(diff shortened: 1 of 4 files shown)


test/elixir_analyzer/test_suite/rpg_character_sheet_test.exs
```diff
@@ -145,6 +145,39 @@ defmodule ElixirAnalyzer.TestSuite.RpgCharacterSheetTest do
           |> IO.gets()
           |> String.trim()
         end
+      end,
+      defmodule RPG.CharacterSheet do
+        import IO
```
angelikatyborska, line 150: Should we also add a test for `newsletter` that uses imports?

angelikatyborska (approved):
I agree it's a good tradeoff. I would always rather let real mistakes slide than show wrong hints.

test/elixir_analyzer/test_suite/rpg_character_sheet_test.exs
jiegillet: Sure! I'll add one tomorrow.

PR #255, opened 2022-02-16
Remove `status` and `meta` from `feature`

jiegillet:
In this PR, I removed all mentions of `status` and `meta` from `feature`.

Keeping the metadata from the solutions would make the solution whitespace dependent, and skipping test was probably only useful during early development. 

Both of these were completely unused.

### DIFF
--- a/docs/step-02/step-02.md
+++ b/docs/step-02/step-02.md
@@ -46,17 +46,17 @@ Then we can run the elixir analyzer:
 Compiling 2 files (.ex)
 Generated escript elixir_analyzer with MIX_ENV=dev
 
-> ./elixir_analyzer --analyze-file "<path>/02-example-module.ex:Example"
+> ./elixir_analyzer example-slug path/to/example/folder path/to/output 
 undefined analysis ... Completed!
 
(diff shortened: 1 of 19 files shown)


neenjaw (approved):
I feel pressured to respond -- in the previous pr I saw that you had plans to wait for my comment before making this decision, but then here is a pr to remove it.

It's alright that you have made this decision, but it could be suggested more directly next time.

I don't want to waste your initiative and work if you feel this is the right move to make, but I haven't really had the head space to think about if we should make this change in terms of long term possibilities that having them keeps.

The changes themselves appear reasonable within the bounds of the proposed pr.

jiegillet:
Hi @neenjaw, I apologize for pressuring you. I did indeed mention that I would wait for your comment, and I should have done that.
If you want more time to think about it, please take as much as you need, there is nothing urgent here.

neenjaw:
Appreciate your reply. As I mentioned, I'm not interested in wasting your work or initiative and at least in this case I have no strong feeling either way about this change at this time. 

Please feel free to proceed.

angelikatyborska (approved):

PR #257, opened 2022-02-26
Add coverage badge on README

jiegillet:
This PR adds a coverage badge on the README (but only on push to main). [Example here on my fork](https://github.com/jiegillet/elixir-analyzer) (coverage is 99% because I was testing stuff, it will be 100% here).

We can [check here](https://coveralls.io/github/exercism/elixir-analyzer) the settings of the badge, history of coverage, etc. At least I can, you may have to add the repo first.

[There are actions](https://github.com/marketplace/actions/coveralls-github-action) to add comments on PR and all sorts of stuff, but I don't think we need that?

Also, running `mix test --cover` is taken over by coveralls and shows more details. Running `MIX_ENV=test mix coveralls.html` is super useful when working on coverage since it produces a `cover/excoveralls.html` that has all the source code and highlights which lines are covered or not.


angelikatyborska (approved):

angelikatyborska:
Thank you for this! It's really cool

jiegillet:
> I have access in theory, but in practice it says there's no history. I think it's because the only jobs were run on your own fork 🤔

Yep, there is something now!

> > [There are actions](https://github.com/marketplace/actions/coveralls-github-action) to add comments on PR and all sorts of stuff, but I don't think we need that?
> 
> I actually really like those comments. They are a nice prompt to add more tests to your PR (but I wouldn't configure any blocking checks that the coverage must not decrease to merge the PR).

We can try it out then.
The thing is, the exercise checks are too declarative (using `feature` and such) and don't appear on the list of things to cover. I tried adding one with no unit test and even a failing test (no slug in config) and coverage was still 100%. So we need to make sure that it doesn't send the wrong message that not adding tests is OK.
It would be neat for PRs that change actual code.

PR #258, opened 2022-02-26
Fix org in coveralls badge

angelikatyborska:
I was too fast on that merge button 😝


jiegillet:
Oops! Sorry about that :p

PR #260, opened 2022-03-02
Update dependabot label size to tiny

jiegillet:


jiegillet:
Mmh.. Before I do the other repos, is this the best way to do things? Or should it be a global update for all exercism repos that have dependabot?

PR #269, opened 2022-03-27
Update `boutique-inventory` exemplar in tests

jiegillet:
I accidentally merged a dependabot PR that ended up failing the tests because the `boutique-inventory` exemplar was updated in `elixir`.

This is a quick fix to address my mistake, but feel free to add the new analyzer checks on top.


angelikatyborska:
I'll do a separate PR with more checks soon 🙂 thanks!

angelikatyborska (approved):

PR #285, opened 2022-04-27
Add exception `%{}` to function capture common check

### LINKED ISSUE #282: Suggests that arty zero anon functions are converted to short form.
Here is the sample code (from Community Garden):

```
  def start() do
    Agent.start(fn -> %{} end)
  end
```

This is the warning generated:

```
When passing an existing function as an argument, make sure to capture it instead of creating a new anonymous function.

# preferred: using capture notation

&%{}/0

# discouraged: creating new anonymous functions

fn -> %{}() end
```

jiegillet:
Closes #282.

### DIFF
--- a/test/elixir_analyzer/exercise_test/common_checks/function_capture_test.exs
+++ b/test/elixir_analyzer/exercise_test/common_checks/function_capture_test.exs
@@ -48,6 +48,7 @@ defmodule ElixirAnalyzer.ExerciseTest.CommonChecks.FunctionCaptureTest do
             |> Enum.map(fn x -> <<x>> end)
             |> Enum.reduce([], fn a, b, c -> {a, b, c} end)
             |> Enum.reduce([], &{&1, &2, &2})
+            |> Enum.map(fn -> %{} end)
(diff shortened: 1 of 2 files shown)


test/elixir_analyzer/exercise_test/common_checks/function_capture_test.exs
```diff
@@ -48,6 +48,7 @@ defmodule ElixirAnalyzer.ExerciseTest.CommonChecks.FunctionCaptureTest do
             |> Enum.map(fn x -> <<x>> end)
             |> Enum.reduce([], fn a, b, c -> {a, b, c} end)
             |> Enum.reduce([], &{&1, &2, &2})
+            |> Enum.map(fn -> %{} end)
```
jiegillet, line 51: I was trying to find ways to have some kind of valid `fn ??? -> %{???}` notation, but I couldn't find any.

jiegillet:
Run CI again after #286 is merged.

angelikatyborska (approved):

PR #287, opened 2022-05-02
Bug fixes and improvements to function annotation order common check

### LINKED ISSUE #284: Warns about order of doc and spec attributes, but I can't find the error
Here is the code:

```elixir
defmodule School do
  @moduledoc """
  Simulate students in a school.
 
  Each student is in a grade.
  """
  @type school :: any()
  @doc """
  Create a new, empty school.
  """
  @spec new() :: school
  def new() do
    %{}
  end
  @doc """
  Add a student to a particular grade in school.
  """
  @spec add(school, String.t(), integer) :: {:ok | :error, school}
  def add(school, name, grade) do
    if school |> Map.get(name) != nil do
      {:error, school}
    else
      {:ok, school |> Map.put(name, grade)}
    end
  end
  @doc """
  Return the names of the students in a particular grade, sorted alphabetically.
  """
  @spec grade(school, integer) :: [String.t()]
  def grade(school, grade) do
    (for {k, v} <- school, v == grade, do: k) |> Enum.sort()
  end
  @doc """
  Return the names of all the students in the school sorted by grade and name.
  """
  @spec roster(school) :: [String.t()]
  def roster(school) do
    (for {k, v} <- school, do: {k, v} )
    |> Enum.sort(School.Sorting)
    |> Enum.map(&elem(&1, 0))
  end
  defmodule Sorting do
    @doc """
    Provides a compare function
    """
    @spec compare(first :: {String.t(), integer}, second :: {String.t(), integer}) :: :gt | :lt | :eq
    def compare({name, grade}, {name2, grade2}) do
      cond do
        grade > grade2 -> :gt
        grade < grade2 -> :lt
        name > name2 -> :gt
        name < name2 -> :lt
        true -> :eq
      end
    end
  end
end
```

Here is the warning: 
```
Developers can choose the order of the @doc and @spec modules attributes, but the Elixir community convention is to use @doc first and @spec next to the function.

Example:

@doc """
This function counts the number elements in a list
"""
@spec count(list(any())) :: integer()
def count(list), do: length(list)
```

jiegillet:
Closes #284.

The bug fix for #284 was quite simple, but I kept going, because I wasn't totally satisfied with the quality of this common check. 

I ended up modifying the code a lot, but I made sure to always add failing tests before working on a feature, so I recommend going through it commit by commit to keep track of the context. 

In a nutshell, here are some issues that this PR fixes:
- List of definition was not getting reversed at transition between modules/submodules (bug in #284)
- AST traversal changed the AST, causing to skip modules with a single function
- AST traversal didn't handle transitions between modules/submodules correctly
- code didn't handle overloaded specs
- code handled private functions/macros and guards, which are not supposed to have public docs and specs (I guess)
- there were other issues due to my own modifications that I fixed at the end

### DIFF
--- a/lib/elixir_analyzer/exercise_test/common_checks/function_annotation_order.ex
+++ b/lib/elixir_analyzer/exercise_test/common_checks/function_annotation_order.ex
@@ -9,7 +9,7 @@ defmodule ElixirAnalyzer.ExerciseTest.CommonChecks.FunctionAnnotationOrder do
   Common check to be run on every single solution.
   """
 
-  @def_ops [:def, :defp, :defmacro, :defmacrop, :defguard, :defguardp]
+  @def_ops [:def, :defmacro]
 
   def run(ast) do
     acc = %{module: [], definitions: %{}}
@@ -124,9 +124,13 @@ defmodule ElixirAnalyzer.ExerciseTest.CommonChecks.FunctionAnnotationOrder do
     Enum.uniq(operations) not in [
       [],
       [:def],
+      [:defmacro],
       [:spec, :def],
+      [:spec, :defmacro],
       [:doc, :def],
-      [:doc, :spec, :def]
+      [:doc, :defmacro],
+      [:doc, :spec, :def],
+      [:doc, :spec, :defmacro]
     ]
   end
 end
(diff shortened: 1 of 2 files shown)


angelikatyborska (approved):
No notes, looks great, I trust the tests

Issue #283, opened 2022-04-27
Warns about a compare function used in a sort


chriseyre2000:
Here is the code:

```
defmodule School do
  @moduledoc """
  Simulate students in a school.
 
  Each student is in a grade.
  """
  @type school :: any()
  @doc """
  Create a new, empty school.
  """
  @spec new() :: school
  def new() do
    %{}
  end
  @doc """
  Add a student to a particular grade in school.
  """
  @spec add(school, String.t(), integer) :: {:ok | :error, school}
  def add(school, name, grade) do
    if school |> Map.get(name) != nil do
      {:error, school}
    else
      {:ok, school |> Map.put(name, grade)}
    end
  end
  @doc """
  Return the names of the students in a particular grade, sorted alphabetically.
  """
  @spec grade(school, integer) :: [String.t()]
  def grade(school, grade) do
    (for {k, v} <- school, v == grade, do: k) |> Enum.sort()
  end
  @doc """
  Return the names of all the students in the school sorted by grade and name.
  """
  @spec roster(school) :: [String.t()]
  def roster(school) do
    (for {k, v} <- school, do: {k, v} ) 
    |> Enum.sort(School)
    |> Enum.map(&elem(&1, 0))
  end
  def compare({name, grade}, {name2, grade2}) do
    cond do
        grade > grade2 -> :gt
        grade < grade2 -> :lt
        name > name2 -> :gt
        name < name2 -> :lt
        true -> :eq
      end
  end
end
```

It raises the following warning:

```
The public interface of a codebase should be carefully considered, as each public function increases long term maintenance costs and makes refactoring harder.

In every Exercism exercise, the public interface is defined by the test suite. If a function is being tested, it needs to the public. All other functions should be private, including functions with the same name but different arities.

# you defined the following public function

def compare(_, _)

# make it private like this

defp compare(_, _)

```

jiegillet:
Just to be clear, your `compare/2` is overriding the default behavior of comparison of a a 2-tuple?
I had no idea you could do that.
I also have no idea how to tackle this corner case...

chriseyre2000:
It's one of the versions of Enum.sort, provide a module name with a compare function implemented in it.
I am one of the crazy people who reads the documentation.
It was intended to allow a struct to define how it is sorted, but here I am using it for a 2-tuple.

The lazy option would be to permit compare/2 to be defined if Enum.sort/2 is used.

chriseyre2000:
[...]

Issue #276, opened 2022-04-11
TakeANumberDeluxe always returns compliation warnings about unknown module State


angelikatyborska:
Note this analysis result for the exemplar solution:
![Screenshot 2022-04-11 at 21-10-42 Exercism](https://user-images.githubusercontent.com/7852553/162812512-e6d89f7b-47c7-42b3-9653-a35c4e522b2e.png)

angelikatyborska:
I think the solution has to be to prepend/append editor files to the code before it gets compiled for the sake of compilation warnings 🤔 

This doesn't affect dancing-dots because there the other module only gets used for type definitions.

angelikatyborska:
As a quick fix, we think we could also use `use ElixirAnalyzer.ExerciseTest, suppress_tests: [Constants.solution_compiler_warnings()]` on this exercise. Not ideal, but better than false-positives.

angelikatyborska:
> For the compiler warning, we might have to switch to Kernel.ParallelCompiler.compile/2 instead.

It does sound like what we need: "Those files are compiled in parallel and can automatically detect dependencies between them.".

I think there are multiple topics here, multiple levels of complexity...
1. Support more than 1 solution file, read from the config. All of the analyzer checks should run on all of them because all of them must be edited by the student. So far we don't have such an exercise, but dancing-dots could have been like this (you need to write 3 modules).
2. Support editor files. This I think matters mostly for compilation warnings. Editor files aren't supposed to be edited by students, so any analyzer comments coming from them would be our fault. But they need to be compiled to correctly report compilation warnings for solution files.
3. Support arbitrary files. This one is the trickiest but could also simultaneously cover the other two cases. There are definitely analyzer checks that would be great to run on arbitrary files: variable naming, compilation warnings, unless/else etc. Mostly the common checks, not exercise specific checks that make a lot of assumptions about the solution. But should we tell apart solution files and editor files from all the `lib/.*` files and give them extra treatment? E.g. to prevent reporting compilation errors from editor files that we ourselves wrote. 

> I've always thought it would be best for the analyzer to read and concat all of the .ex files in lib into one instead of just reading the solution file.

For the compilation warnings, we need to retain file names and line numbers. For everything else, it might work. I guess we should just try something and see which problems arise 🙃

jiegillet:
[...]

PR #289, opened 2022-05-15
Allow public compare/2 functions when used in Enum.sort

angelikatyborska:
Closes https://github.com/exercism/elixir-analyzer/issues/283

@jiegillet I hope this doesn't conflict with your work. You said you wanted to tackle this issue after https://github.com/exercism/elixir-analyzer/issues/276 and I figured I can help out because you didn't start yet, right?

### DIFF
--- a/lib/elixir_analyzer/exercise_test/common_checks/private_helper_functions.ex
+++ b/lib/elixir_analyzer/exercise_test/common_checks/private_helper_functions.ex
@@ -17,10 +17,15 @@ defmodule ElixirAnalyzer.ExerciseTest.CommonChecks.PrivateHelperFunctions do
     {_, %{definitions: code_definitions}} =
       Macro.traverse(code_ast, acc, &annotate/2, &find_definition/2)
 
+    # Some functions have to be public even though they don't appear in the exemploid.
+    # That's because they can be indirectly used with dynamic dispatch by the standard library.
+    {_, %{exceptions: exceptions}} =
+      Macro.traverse(code_ast, %{module: [], exceptions: []}, &annotate/2, &find_exceptions/2)
+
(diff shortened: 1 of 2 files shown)


lib/elixir_analyzer/exercise_test/common_checks/private_helper_functions.ex
```diff
@@ -111,4 +117,79 @@ defmodule ElixirAnalyzer.ExerciseTest.CommonChecks.PrivateHelperFunctions do
...
+  defp find_enum_sort_exceptions(sorter_arg, %{module: module, exceptions: exceptions}) do
+    case sorter_arg do
+      order when order in [:asc, :desc] ->
+        exceptions
+
+      {:__aliases__, _, module} when is_list(module) ->
+        [{[module], :def, :compare, 2} | exceptions]
+
+      {:__MODULE__, _, _} when is_list(module) ->
+        [{module, :def, :compare, 2} | exceptions]
```
jiegillet, line 183: The use of `module` is a bit confusing in some of these cases, because this one refers to the `module` from the function argument, but in other cases, they are the module from `sorter_arg`. Those two are very close but not quite the same, so I would suggest a different name. Maybe `arg_module`?

jiegillet (approved):
I had not started working on this, so it's perfect :)

I left one tiny comment about improving readability, otherwise everything looks great!

angelikatyborska:
I noticed that my code doesn't handle aliasing... I will merge it anyway now, and we can improve in a separate PR.

Do you know if it would be possible to somehow reuse the alias tracking that's defined in `ElixirAnalyzer.ExerciseTest.AssertCall.Compiler`? It would be a shame to have to write the same thing again.

Issue #278, opened 2022-04-12
Extend TakeANumberDeluxe `@impl` check to check all callbacks


angelikatyborska:
In https://github.com/exercism/elixir-analyzer/pull/277, we added a simple check that detects if `@impl GenServer` is used anywhere in the code at least once. Improve this check so that it actually verifies that all callback definitions are annotated: `init/2`, `handle_call/3`, `handle_cast/2`, `handle_info/2`.

PR #290, opened 2022-05-22
Add exhaustive @impl GenServer check

angelikatyborska:
Closes https://github.com/exercism/elixir-analyzer/issues/278

### DIFF
--- a/lib/elixir_analyzer/test_suite/take_a_number_deluxe.ex
+++ b/lib/elixir_analyzer/test_suite/take_a_number_deluxe.ex
@@ -4,6 +4,7 @@ defmodule ElixirAnalyzer.TestSuite.TakeANumberDeluxe do
   """
 
   alias ElixirAnalyzer.Constants
+  alias ElixirAnalyzer.Source
 
   use ElixirAnalyzer.ExerciseTest,
     # this is temporary until we include editor files in compilation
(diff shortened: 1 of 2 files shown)


lib/elixir_analyzer/test_suite/take_a_number_deluxe.ex
```diff
@@ -19,13 +20,37 @@ defmodule ElixirAnalyzer.TestSuite.TakeANumberDeluxe do
...
     end
   end
+
+  @def_ops [:def, :defp, :defmacro, :defmacrop, :defguard, :defguardp]
+  @genserver_callbacks_in_this_exercise [:init, :handle_call, :handle_cast, :handle_info]
+  defp find_defs_and_impls(node, acc) do
+    acc =
+      case node do
+        {:@, _, [{:impl, _, [{:__aliases__, _, [:GenServer]}]}]} ->
+          %{acc | impl?: true}
```
angelikatyborska, line 41: This cannot track aliases given to GenServer

test/elixir_analyzer/test_suite/take_a_number_deluxe_test.exs
```diff
@@ -171,46 +171,203 @@ defmodule ElixirAnalyzer.TestSuite.TakeANumberDeluxeTest do
...
+          @impl GenServer
+          def handle_cast(:reset_state, state) do
+          end
+
+          @impl GenServer
+          def handle_info(:timeout, state) do
+          end
+
+          def handle_info(_, state) do
           end
```
angelikatyborska, line 371: In each example in this list a different callback is missing `@impl GenServer`, in order from the first one to the last one (here)

lib/elixir_analyzer/test_suite/take_a_number_deluxe.ex
jiegillet: I think we can live without this, it's not worth the added complexity.

lib/elixir_analyzer/test_suite/take_a_number_deluxe.ex
```diff
@@ -19,13 +20,37 @@ defmodule ElixirAnalyzer.TestSuite.TakeANumberDeluxe do
...
+
+        {op, _, [{function_name, _, _} | _]} when op in @def_ops ->
+          if function_name in @genserver_callbacks_in_this_exercise and !acc.impl? do
+            %{acc | impl?: false, defs_without_impls: [function_name | acc.defs_without_impls]}
+          else
+            %{acc | impl?: false}
+          end
+
+        _ ->
+          acc
```
jiegillet, line 51: Should we return `%{acc | impl?: false}` here instead?
Otherwise this case wouldn't get flagged:
```elixir
@impl GenServer
defp some_helper_function, do: :something
def handle_call(...)
```

jiegillet (commented):
Looks great, I only have one suggestion.

lib/elixir_analyzer/test_suite/take_a_number_deluxe.ex
jiegillet: I mean, arguably that example might be fine, but there might be worse ones

angelikatyborska: By returning `%{acc | impl?: false}` here, we're saying that `@impl` only counts if it's the last thing before `def`, right? That doesn't sound correct to me 🤔. If I compile this snippet:
```
defmodule FooBar do
  use GenServer

  @impl GenServer
  @doc "some doc"
  @spec this_is_not_a_genserver_callback :: nil
  def this_is_not_a_genserver_callback, do: nil
end
```

I still get a warning, which means that `@impl` counts:
```
warning: got "@impl GenServer" for function this_is_not_a_genserver_callback/0 but this behaviour does not specify such callback. The known callbacks are:
```

jiegillet: OK, I stand corrected.

jiegillet (approved):

Issue #274, opened 2022-04-10
Write analysis for concept exercise `dancing-dots`


angelikatyborska:
Desired checks: https://github.com/exercism/elixir/blob/main/exercises/concept/dancing-dots/.meta/design.md#analyzer

PR #293, opened 2022-05-27
Write analysis for concept exercise `dancing-dots`

angelikatyborska:
Closes https://github.com/exercism/elixir-analyzer/issues/274

Website copy PR: https://github.com/exercism/website-copy/pull/2193

### DIFF
--- a/lib/elixir_analyzer/test_suite/dancing_dots.ex
+++ b/lib/elixir_analyzer/test_suite/dancing_dots.ex
@@ -8,6 +8,20 @@ defmodule ElixirAnalyzer.TestSuite.DancingDots do
 
   use ElixirAnalyzer.ExerciseTest
 
+  feature "DancingDots.Flicker shouldn't reimplement init/1" do
+    type :essential
+    find :none
+    comment Constants.dancing_dots_do_not_reimplement_init()
+
+    form do
+      defmodule DancingDots.Flicker do
+        _block_includes do
+          def init(_ignore), do: _ignore
+        end
+      end
+    end
+  end
+
   check_source "uses @impl DancingDots.Animation" do
     type :actionable
     comment Constants.dancing_dots_annotate_impl_animation()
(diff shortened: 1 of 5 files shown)


lib/elixir_analyzer/test_suite/dancing_dots.ex
```diff
@@ -0,0 +1,94 @@
...
+    end
+  end
+
+  @def_ops [:def, :defp, :defmacro, :defmacrop, :defguard, :defguardp]
+  @callbacks_in_this_exercise [:init, :handle_frame]
+  @behaviour_module [:DancingDots, :Animation]
+  defp find_aliases(node, acc) do
+    acc =
+      case node do
+        {:alias, _, [{:__aliases__, _, module_name}]} ->
```
jiegillet, line 54: It might be simpler at this stage to only care about aliases of `@behaviour_module`, this way `acc` could simplify to a list (which could also be initialized as `[@behaviour_module]` to simplify the check later).
Unless you intended to reuse the function, but then this is not the right place...

angelikatyborska: Sure: [`f02db59` (#293)](https://github.com/exercism/elixir-analyzer/pull/293/commits/f02db5922c3595826adbaf6e187ec378ce238e77)

test/elixir_analyzer/test_suite/dancing_dots_test.exs
```diff
@@ -0,0 +1,270 @@
+defmodule ElixirAnalyzer.TestSuite.DancingDotsTest do
+  use ElixirAnalyzer.ExerciseTestCase,
+    exercise_test_module: ElixirAnalyzer.TestSuite.DancingDots
+
+  test_exercise_analysis "example solution",
+    comments: [ElixirAnalyzer.Constants.solution_same_as_exemplar()] do
+    ~S"""
```
jiegillet, line 7: I didn't know we couldn't define more than one module at a time... Is that worth looking into?

jiegillet (approved):
Very nice :)

test/elixir_analyzer/test_suite/dancing_dots_test.exs
angelikatyborska: Yeah, it doesn't work. I assume it would be possible to find the right addition to this snippet that would make it possible: https://github.com/exercism/elixir-analyzer/blob/0a1bd66587808f72cec0ef7404fc6708da0058ad/test/support/exercise_test_case.ex#L86-L90

But I don't feel a strong need to work on it at the moment 😇

Issue #80, opened 2021-07-07
Write analysis for concept exercise `basketball-website`


angelikatyborska:
Desired checks:
https://github.com/exercism/elixir/blob/9774308bb6b898df6ecbde64c5c6e48c651870ae/exercises/concept/basketball-website/.meta/design.md#L40-L46

Except for what's in the `design.md` file, also add a check to discourage handling `nil` data explicitly because `nil` already implements the access behavior and returns `nil` for anything. Below I'm giving a few community solution examples with this problem.

After analyzing them, I think this check could be as broad as putting `nil` explicitly anywhere in the code, doesn't matter where. There are no usages of `nil` in a correct solution that wouldn't be redundant.

```elixir
  def extract_from_path(data, path) do
    path_names = String.split(path, ".")
    do_extract(data, path_names)
  end
  defp do_extract(:nil, _) do
    :nil
  end
  defp do_extract(data, []) do
      data
  end
  defp do_extract(data, [h|t]) do
    do_extract(data[h], t)
  end
```

```elixir
  def extract_from_path(data, path) do
    extract_from_split_path(data, String.split(path, "."))
  end
  defp extract_from_split_path(data, []), do: data
  defp extract_from_split_path(data, [k | rest_path]) do
    case data do
      %{^k => v} -> extract_from_split_path(v, rest_path)
      _ -> nil
    end
  end
```
```elixir
  def extract_from_path(data, path) do
    do_extract_from_path(data, String.split(path, ~r/\./))
  end
  defp do_extract_from_path(nil, _path), do: nil
  defp do_extract_from_path(data, []), do: data
  defp do_extract_from_path(data, [top | tail]) do
    do_extract_from_path(data[top], tail)
  end
```

PS: I am not 100% sure if the first two checks can be implemented easily or even how to come up with correct solutions that would violate them. Feel free to ignore those two if you don't figure it out either 🙂 (suggested checks in `design.md` files were written a long time ago before students started solving the exercises)

> 1. That structs aren't used.
> 2. That the static access operator isn't used.

PR #294, opened 2022-05-28
Add analyzer check for concept exercise basketball-website

angelikatyborska:
Closes https://github.com/exercism/elixir-analyzer/issues/80

It doesn't cover those two cases mentioned in `design.md` because I couldn't figure out how features that require keys to be known at compile time could possibly be used in this exercise:

> - That structs aren't used.
> - That the static access operator isn't used.

Website copy PR: https://github.com/exercism/website-copy/pull/2194

### DIFF
--- a/test/elixir_analyzer/test_suite/basketball_website_test.exs
+++ b/test/elixir_analyzer/test_suite/basketball_website_test.exs
@@ -2,23 +2,45 @@ defmodule ElixirAnalyzer.ExerciseTest.BasketballWebsiteTest do
   use ElixirAnalyzer.ExerciseTestCase,
     exercise_test_module: ElixirAnalyzer.TestSuite.BasketballWebsite
 
-  test_exercise_analysis "example solution",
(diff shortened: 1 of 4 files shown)


test/elixir_analyzer/test_suite/basketball_website_test.exs
```diff
@@ -0,0 +1,124 @@
...
+          paths = String.split(path, ".", trim: true)
+          get_in(data, paths)
+        end
+      end
+    end
+
+    test_exercise_analysis "using Enum is also fine", comments: [] do
+      defmodule BasketballWebsite do
+        def extract_from_path(data, path) do
+          Enum.reduce(String.split(path, "."), data, fn k, acc -> acc[k] end)
```
jiegillet, line 50: Oh, I like this one!

jiegillet (approved):
Looks great.
I also couldn't find a way to use structs or the access notation. I guess that was the reason it took so long to get to this one, and it wasn't even justified 🙈

PR #300, opened 2022-06-16
Add check to prevent calls to defdelegate

### LINKED ISSUE #296: Accept defdelegates as valid equivalents of function calls in `assert_call`
Hi, I'd like to start off with saying I love this tool and the exercises.
Now I've been using Elixir for a couple of years already and I'm kinda trying to work my way through the problems as fast as possible, and I implemented the "High Score' problem using `defdelegate`. Now I understand this isn't the point of the exercise, but the analysis remarks I got weren't correct.

This is the code:
```
defmodule HighScore do
  @default_score 0
  defdelegate new, to: Map
  defdelegate add_player(scores, name, score \\ @default_score), to: Map, as: :put
  defdelegate remove_player(scores, name), to: Map, as: :delete
  defdelegate reset_score(scores, name), to: __MODULE__, as: :add_player
  def update_score(scores, name, score), do: Map.update(scores, name, score, fn x -> score + x end)
  defdelegate get_players(scores), to: Map, as: :keys
end
```

This was the feedback of the analysis:
> Essential
Use a [module attribute](https://elixir-lang.org/getting-started/module-attributes.html#as-constants) in add_player/3 and reset_score/2 as a constant for the initial score.
Recommended
Use a [default argument](https://elixir-lang.org/getting-started/modules-and-functions.html#default-arguments) in the add_player/3 function to handle a lack of initial score. Use the module attribute with the default initial score.

Even though the `defdelegate`s are probably cheating, I am using default arguments and a module attribute, so IMO the analysis is not correct.
Maybe it's an option to instead add feedback to not use `defdelegate` as well, as it defeats the purpose of the exercise.

TBK145:
Closes #296

### DIFF
--- a/lib/elixir_analyzer/constants.ex
+++ b/lib/elixir_analyzer/constants.ex
@@ -42,6 +42,7 @@ defmodule ElixirAnalyzer.Constants do
     solution_unless_with_else: "elixir.solution.unless_with_else",
     solution_use_function_capture: "elixir.solution.use_function_capture",
     solution_deprecated_random_module: "elixir.solution.deprecated_random_module",
+    solution_defdelegate: "elixir.solution.defdelegate",
 
     # Concept exercises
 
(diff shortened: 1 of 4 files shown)


angelikatyborska:
Superseded by https://github.com/exercism/elixir-analyzer/pull/363

Issue #305, opened 2022-07-12
Analysis of Basketball Website


altuntasfatih:
I received the following feedback on my answer to the Basketball Website problem:

`"In this exercise, there is no need to handle nil values explicitly. The Access Behaviour is implemented for nil."`

If I do not handle the Nil case, it continues to iterate until the finish path array. However, in my opinion, that is a redundant iteration because the value is already `nil` and never changes. Also, many community solutions are like that; they are iteration redundantly

What is your opinion about that?

Here is my solution also
`
defmodule BasketballWebsite do
  def extract_from_path(data, path) when is_binary(path) do
    [k | tail] = keys(path)
    extract(data[k], tail)
  end

  defp extract(nil, _), do: nil
  defp extract(data, []), do: data
  defp extract(data, [head | tail]), do: extract(data[head], tail)

  def get_in_path(data, path) do
    Kernel.get_in(data, keys(path))
  end

  defp keys(path) do
    String.split(path, ".", trim: true)
  end
end
`

arconautishche:
You could probably simplify your `extract` function and remove the `nil` issue like this:
```elixir
defp extract(data, [only_key]), do: data[only_key]
defp extract(data, [head | tail]), do: extract(data[head], tail)
```

Although I'm not sure your original solution isn't just as valid.

altuntasfatih:
@arconautishche, My solution is correct, but I am trying to point out another thing.

The feedback says you should not handle the `nil` case because the `Access` module already handles it like in your simplified version.
In your simplified version( also many other community solutions), the code iterates all path items even when data is `nill.`

In brief, the above feedback directs people to a solution that is not very proper.

angelikatyborska:
@altuntasfatih You're absolutely right. I didn't notice that the pattern matching on `nil` has a purpose, at a first glance it looked unnecessary. We should remove that piece of feedback from the analyzer. Would you want to open a PR for that?

altuntasfatih:
@angelikatyborska I would like to open pr for that

PR #309, opened 2022-07-15
Remove 'no explicit nil handling feedback' from basketball_website

altuntasfatih:
As we discussed this [issue](https://github.com/exercism/elixir-analyzer/issues/305), I have just removed `nil case feedback` from the basketball_website analyzer.

Resolves https://github.com/exercism/elixir-analyzer/issues/305

### DIFF
--- a/test/elixir_analyzer/test_suite/basketball_website_test.exs
+++ b/test/elixir_analyzer/test_suite/basketball_website_test.exs
@@ -59,7 +59,7 @@ defmodule ElixirAnalyzer.ExerciseTest.BasketballWebsiteTest do
 
   describe "using Map" do
     test_exercise_analysis "there should be no usage of the map module",
-      comments_include: [ElixirAnalyzer.Constants.basketball_website_no_explicit_nil()] do
+      comments_include: [ElixirAnalyzer.Constants.basketball_website_no_map()] do
       defmodule BasketballWebsite do
         def extract_from_path(data, path) do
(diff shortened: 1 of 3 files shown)


test/elixir_analyzer/test_suite/basketball_website_test.exs
```diff
@@ -57,25 +57,6 @@ defmodule ElixirAnalyzer.ExerciseTest.BasketballWebsiteTest do
...
-        defp do_extract(data, [path | next]) do
-          do_extract(data[path], next)
-        end
-      end
-    end
-  end
-
   describe "using Map" do
     test_exercise_analysis "there should be no usage of the map module",
       comments_include: [ElixirAnalyzer.Constants.basketball_website_no_explicit_nil()] do
```
angelikatyborska, line 62: Oops, I see we made a mistake in this test in the past. It shouldn't use the comment about nil. In fact, after this PR, the "constant" `ElixirAnalyzer.Constants.basketball_website_no_explicit_nil()` should be completely gone - can you remove it from the `Constants` module?
```suggestion
      comments_include: [ElixirAnalyzer.Constants.basketball_website_no_map()] do
```

altuntasfatih: I have just amended by your feedback.

angelikatyborska (approved):
Thanks a lot 💜

Issue #316, opened 2022-08-13
Take-A-Number-Deluxe @impl required 


louis-cl:
Seems that the analysis requires the `@impl` annotation for all methods.
Where I see [recommendation](https://www.reddit.com/r/elixir/comments/aildm1/impl_for_all_or_just_the_first_implementation/) to only mark the first method of each kind

My solution
```elixir
defmodule TakeANumberDeluxe do
  # Client API

  @spec start_link(keyword()) :: {:ok, pid()} | {:error, atom()}
  def start_link(init_arg) do
    GenServer.start_link(__MODULE__, init_arg)
  end

  @spec report_state(pid()) :: TakeANumberDeluxe.State.t()
  def report_state(machine) do
    GenServer.call(machine, :get_state)
  end

  @spec queue_new_number(pid()) :: {:ok, integer()} | {:error, atom()}
  def queue_new_number(machine) do
    GenServer.call(machine, :new_number)
  end

  @spec serve_next_queued_number(pid(), integer() | nil) :: {:ok, integer()} | {:error, atom()}
  def serve_next_queued_number(machine, priority_number \\ nil) do
    GenServer.call(machine, {:serve, priority_number})
  end

  @spec reset_state(pid()) :: :ok
  def reset_state(machine) do
    GenServer.cast(machine, :reset)
  end

  # Server callbacks
  use GenServer

  @impl GenServer
  def init(opts) do
    timeout = Keyword.get(opts, :auto_shutdown_timeout, :infinity)

    case TakeANumberDeluxe.State.new(opts[:min_number], opts[:max_number], timeout) do
      {:ok, state} -> {:ok, state, timeout}
      {:error, reason} -> {:stop, reason}
    end
  end

  @impl GenServer
  def handle_call(:get_state, _, state) do
    {:reply, state, state, state.auto_shutdown_timeout}
  end

  def handle_call(:new_number, _, state) do
    case TakeANumberDeluxe.State.queue_new_number(state) do
      {:ok, new_number, new_state} ->
        {:reply, {:ok, new_number}, new_state, state.auto_shutdown_timeout}

      err ->
        {:reply, err, state, state.auto_shutdown_timeout}
    end
  end

  def handle_call({:serve, priority_number}, _, state) do
    case TakeANumberDeluxe.State.serve_next_queued_number(state, priority_number) do
      {:ok, next_number, new_state} ->
        {:reply, {:ok, next_number}, new_state, state.auto_shutdown_timeout}

[...]

PR #317, opened 2022-08-13
Require @impl only once per function

angelikatyborska:
Resolves https://github.com/exercism/elixir-analyzer/issues/316


jiegillet (approved):
Looks good.
Warning: I can't seem to run tests locally at the moment, it fails for some ludicrous reason that will probably take me forever to fix...

Issue #323, opened 2022-08-25
Analysis of New Passport


Vagab:
Hey guys. So, given this additional code:
```elixir
def get_new_passport(now, birthday, form) do
  with {:ok, timestamp} <- enter_building(now),
       {:ok, manual} <- validate_no_coffee_break(now),
       counter = manual.(birthday),
       {:ok, checksum} <- stamp_form(timestamp, counter, form),
       number <- get_new_passport_number(timestamp, counter, checksum) do
    {:ok, number}
  end
end

defp validate_no_coffee_break(now) do
  case find_counter_information(now) do
    {:coffee_break, _} -> {:retry, now |> NaiveDateTime.add(15 * 60)}
    ok -> ok
  end
end
```
Analyzer hits me up with:

>This exercise is about learning how to "follow the happy path", with the special form with, which lets you deal with errors at the end. Use with in your implementation of get_new_passport/3.

I understand it wants me to use `else`, but in this case I think it's actually more readable and idiomatic([docs](https://hexdocs.pm/elixir/Kernel.SpecialForms.html#with/1-beware))

angelikatyborska:
Hi! Thanks for reporting. So technically this comment appears because we have a check that `case` shouldn't be used, there's no check that `with` must come with an `else`. If you wrote `validate_no_coffee_break` with a `cond` or `if` instead, the comment wouldn't appear 🙈 Confusing...

Our intention was indeed to make everyone use an `else` clause for the purpose of learning that it exists. The "bad use of else" that you linked mentions:

> Note how we are having to reconstruct the result types of [Path.extname/1](https://hexdocs.pm/elixir/Path.html#extname/1) and [File.exists?/1](https://hexdocs.pm/elixir/File.html#exists?/1) to build error messages.

That doesn't apply to this exercise's desired solution. There are no type assertions on anything returned by `find_counter_information`, just simple pattern matching. Matching on `{:coffee_break, _}` is no different than matching on `{:ok, _}`. Thus I am not convinced that your solution is any more idiomatic  - it's equally valid, but it doesn't fulfill the goal of learning about using `else` with `with`. Of course if you already feel comfortable with Elixir and know you prefer your way, you're more than welcome to ignore the analyzer feedback. We write it mostly to ensure that newcomers actually use the learning exercises to learn what they need to learn 😉.

[...]

PR #324, opened 2022-08-26
Don't care about `case` in new-passport; require `with` with `else`

angelikatyborska:
Resolves https://github.com/exercism/elixir-analyzer/issues/323

As long as `with` was used, the purpose of the exercise was fulfilled. There are perfectly idiomatic solutions that use `case` in a helper, for example:

```elixir
def get_new_passport(now, birthday, form) do
  with {:ok, timestamp} <- enter_building(now),
       {:ok, manual} <- validate_no_coffee_break(now),
       counter = manual.(birthday),
       {:ok, checksum} <- stamp_form(timestamp, counter, form),
       number <- get_new_passport_number(timestamp, counter, checksum) do
    {:ok, number}
  end
end

defp validate_no_coffee_break(now) do
  case find_counter_information(now) do
    {:coffee_break, _} -> {:retry, now |> NaiveDateTime.add(15 * 60)}
    ok -> ok
  end
end
```

### DIFF
--- a/lib/elixir_analyzer/constants.ex
+++ b/lib/elixir_analyzer/constants.ex
@@ -145,6 +145,7 @@ defmodule ElixirAnalyzer.Constants do
 
     # New Passport Comments
     new_passport_use_with: "elixir.new-passport.use_with",
+    new_passport_use_with_else: "elixir.new-passport.use_with_else",
     new_passport_do_not_modify_code: "elixir.new-passport.do_not_modify_code",
 
     # Pacman Rules Comments
--- a/lib/elixir_analyzer/test_suite/new_passport.ex
+++ b/lib/elixir_analyzer/test_suite/new_passport.ex
@@ -4,6 +4,7 @@ defmodule ElixirAnalyzer.TestSuite.NewPassport do
   """
 
   alias ElixirAnalyzer.Constants
+  alias ElixirAnalyzer.Source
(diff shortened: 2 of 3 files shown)


jiegillet:
I think removing this check is not enough, we should replace it with one that checks that `else` is being used (simple enough with a `feature` I think).
Our purpose is to teach about `with` and `else`, so as long as the message is clear about that goal we should keep doing that (and I think it's pretty clear the way it is).
I know the person that opened the ticket suggested otherwise, but we should still nudge the students towards learning new things before they can make better informed decisions.

angelikatyborska:
Added a check for `else`: [`9747a86` (#324)](https://github.com/exercism/elixir-analyzer/pull/324/commits/9747a86993345fc8c9744272dc8c917ae13012a4) but it will only work if `with`/`else` is used directly in the desired function. I think if people added a helper function there, it would be a mistake by itself (too much boilerplate), so I'm fine with that drawback.

Website copy PR: https://github.com/exercism/website-copy/pull/2219

jiegillet (approved):
Looks good!

I also found that
```elixir
    form do
      with _ignore, _ignore, _ignore, _ignore do
        _ignore
      else
        _ignore
      end
    end
```
works, but I guess the 4 `_ignore` in a row makes it less robust to variations, your version works best.

Issue #318, opened 2022-08-18
Common check for @doc and @spec order doesn't handle private functions or private macros correctly


petros:
I was working on the Name Badge exercise, where I decided to have a `@spec` for two private functions. In my editor, locally, I initially had a `@doc` and a `@spec` for those functions. ElixirLS informed me that the `@doc` strings will be stripped out because those are private functions. I decided to remove the `@doc` strings and only keep the `@spec`s.

<details>
<summary>The code (SPOILER)</summary>

```elixir
defmodule NameBadge do
  @separator " - "

  @doc """
  Take an id, a name, and a department and format a string
  that can be printed on a badge. It handles not having an ID which
  is the case for new employees. And it recognizes and onwer when
  no department is passed.
  """
  @spec print(integer() | nil, String.t(), String.t() | nil) :: String.t()
  def print(id, name, department) do
    format_id(id) <>
      name <>
      format_department(department)
  end

  @spec format_id(integer() | nil) :: String.t()
  defp format_id(id) do
    if is_nil(id), do: "", else: "[#{id}]" <> @separator
  end

  @spec format_department(String.t() | nil) :: String.t()
  defp format_department(department) do
    if is_nil(department),
      do: @separator <> "OWNER",
      else: @separator <> String.upcase(department)
  end
end
```

</details>

When I submitted my iteration, I got the following from the Elixir Analyzer:

> Developers can choose the order of the `@doc` and `@spec` modules attributes, but the Elixir community convention is to use `@doc` first and `@spec` next to the function.

I do understand that, but why would I get it for a function that only has a `@spec`, and not a `@doc` string?

I would understand it if the recommendation was to also have a `@doc` string. But referring to the order of something that does not exist is confusing.

angelikatyborska:
Thank you for the bug report! There's definitely something wrong. The check for the order of `@doc` and `@spec` seems not to handle private functions very well.

This code produces a comment about a wrong order when it shouldn't:
```elixir
defmodule Test do
  @spec x()
  def x()

  @spec y()
  defp y()
end
```

Teilzeitdenker:
Well, I think one only has to change the last function in 
[/elixir-analyzer/tree/main/lib/elixir_analyzer/exercise_test/common_checks/function_annotation_order.ex](https://github.com/exercism/elixir-analyzer/blob/main/lib/elixir_analyzer/exercise_test/common_checks/function_annotation_order.ex)
to
```elixir
[...]

PR #332, opened 2022-09-13
fixed issue #318, order of @doc and @spec in private functions or macros

Teilzeitdenker:
By just adding three lines in the `check_wrong_order()` function, one is able to allow for all combinations with `:defp` and `:defmacrop` at once. Since the nodes corresponding to different functions are chunked and merged accordingly, there is no harm done by adding these to the allowed combinations. 
The three examples from Angelika that produced comments before now run smoothly. 
A wrong order of `@spec - @doc - def(p)/defmacro(p)` still doesn't pass. 
`mix test` also passes completely.

I think one may leave out the `@def_ops` completely, such that the `check_wrong_order()` function just reads something like
```elixir
defp check_wrong_order(%{operations: operations}) do
    Enum.uniq(operations) not in [
      [],
      [:spec],
      [:doc],
      [:doc, :spec],
    ]
  end
end
```
but I didn't try it.

I'm not sure if I should write corresponding tests for the new behaviour and if so, in which file I should do that. 
What's your opinion?

### DIFF
--- a/lib/elixir_analyzer/exercise_test/common_checks/function_annotation_order.ex
+++ b/lib/elixir_analyzer/exercise_test/common_checks/function_annotation_order.ex
@@ -21,52 +21,52 @@ defmodule ElixirAnalyzer.ExerciseTest.CommonChecks.FunctionAnnotationOrder do
     |> check_errors()
   end
 
-  def enter_node({:defmodule, _, [{:__aliases__, _, aliases}, _]} = ast, acc) do
+  defp enter_node({:defmodule, _, [{:__aliases__, _, aliases}, _]} = ast, acc) do
     module = [aliases | acc.module]
     definitions = Map.put(acc.definitions, module, [])
     {ast, %{module: module, definitions: definitions}}
   end
 
-  def enter_node({:@, _, [{:spec, _, [{:"::", _, [{fn_name, _, _} | _]}]} | _]} = ast, acc) do
+  defp enter_node({:@, _, [{:spec, _, [{:"::", _, [{fn_name, _, _} | _]}]} | _]} = ast, acc) do
     definitions = Map.update!(acc.definitions, acc.module, &[{:spec, fn_name} | &1])
     {ast, %{acc | definitions: definitions}}
   end
 
-  def enter_node({:@, _, [{:spec, _, [{fn_name, _, _}]}]} = ast, acc) do
+  defp enter_node({:@, _, [{:spec, _, [{fn_name, _, _}]}]} = ast, acc) do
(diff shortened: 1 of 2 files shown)


jiegillet:
Yes, please tests so we can see the behavior difference.
Ideally those would be tests that initially fail before making changes to the code.

Teilzeitdenker:
Ok, I wrote some tests that would't pass without the fix.
Please review the changes.

jiegillet (commented):
(sorry for the slow response)

Thank you for adding the tests, they do indeed address the opened issue, but in a shallow way and I think we can go further. The core problem with the current situation is that private functions are ignored.
As @angelikatyborska mentioned [in this comment](https://github.com/exercism/elixir-analyzer/issues/318#issuecomment-1241621541), a stronger starting point would be to include `defp`, `defmacrop`, `defguard` and `defguardp` to the mix. This way we could also find the wrong order with these cases.
Then, something more clever could be done with `check_wrong_order` so we don't have to manually list all the possible combinations, but group them as `[:spec, :def_op]` or something.

Before that, we should add failing tests where using private functions with a wrong order of `@doc` and `@spec` actually does raise an analyzer comment.

The reason I didn't originally include the lines you added to `check_wrong_order` is that if you defined modules in this way without a function following them, you get a compilation warning or error. However it is a good idea not to raise a confusing analyzer message in those cases, so I we can leave them (but add test that specify that behavior).

Teilzeitdenker:
Yes, I see. It was just the shortest (and probably dirtiest) fix for the (in my experience with the track exercises) most common bug, when defining a private function with a `@spec` always triggers a misleading (and after a while a little annoying) analyzer comment. 

I will go back to the start (if I'm able to roll back git to the initial point) and try to apply your suggested changes (if I can't find a way, I'll admit it within two weeks or so, I'm still learning, so it's still a bit hard for me to manipulate the AST). 

By the way: Should I put any new tests in a separate pull request or does it suffice to put them in a separate commit, such that it is possible for you to run them first on the old and then on some new version from a later commit?

jiegillet:
You don't need to roll back anything, you can take it from this point on. The tests are great and should stay, the changes in `check_wrong_order` can stay as well (if they are still necessary).
You should keep going in this PR and commit whatever changes, with git it's always possible to jump around in the timeline, even for specific files if necessary.

Take your time, and do ask for help if you need some. I think the changes I'm talking about won't require manipulating the AST too much, just tweaking what we already have.

jiegillet:
@Teilzeitdenker do you mind if I take this PR over? 
If you don't, I'll jump in in a couple o days.

Teilzeitdenker:
No, I don't mind...

jiegillet:
Closing in favor of #342

PR #348, opened 2022-11-24
Be consistent when invoking bash in scripts

dmarcoux:
This aligns with the other scripts which are already invoking bash with `/usr/bin/env bash`.


jiegillet (approved):
Thanks

Issue #353, opened 2022-12-04
FileSniffer analyzer spurious comments?


rongcuid:
The following is my solution:

```elixir
defmodule FileSniffer do
  def type_from_extension("exe"), do: "application/octet-stream"
  def type_from_extension("bmp"), do: "image/bmp"
  def type_from_extension("png"), do: "image/png"
  def type_from_extension("jpg"), do: "image/jpg"
  def type_from_extension("gif"), do: "image/gif"
  def type_from_extension(_), do: nil

  def type_from_binary(<<0x42::8,0x4D::8,_::binary>>), do: "image/bmp"
  def type_from_binary(<<0x47::8,0x49::8,0x46::8,_::binary>>), do: "image/gif"
  def type_from_binary(<<0x89::8,0x50::8,0x4E::8,0x47::8,0x0D::8,0x0A::8,0x1A::8,0x0A::8,_::binary>>), do: "image/png"
  def type_from_binary(<<0xFF::8,0xD8::8,0xFF::8,_::binary>>), do: "image/jpg"
  def type_from_binary(<<0x7F::8,0x45::8,0x4C::8,0x46::8,_::binary>>), do: "application/octet-stream"
  def type_from_binary(_), do: nil

  def verify(file_binary, extension) do
    ty = type_from_extension(extension)
    if ty == type_from_binary(file_binary) do
      {:ok, ty}
    else
      {:error, "Warning, file format and file extension do not match."}
    end
  end
end
```

And I get the following comment:

    Even though there could be other valid ways to solve this exercise, for the sake of mastering bitstrings please match the first few bytes of the argument of type_from_binary/1 with the special form <<>>.

I certainly used only the special form, except for the last clause.

PR #354, opened 2022-12-04
Allow ::8 or ::size(8) in file-sniffer pattern matching

angelikatyborska:
Resolves https://github.com/exercism/elixir-analyzer/issues/353

This is very not-dry 😬 is it good enough or should I look into doing more meta programming with `check_source`?

Also, I am not entirely sure if no size, `8`, and `size(8)` are the only three valid options. That's the only ones I could think of, but maybe there's more?


jiegillet (approved):
It's fine, it's verbose, but very straightforward.
`check_source` always carries a bigger risk because tests are very limited.
Now if we realize there are 7 other types of syntax we didn't cover, we can revisit that.

PR #363, opened 2023-02-10
Accept defdelegates as valid equivalents of function calls in assert_call

angelikatyborska:
Resolves https://github.com/exercism/elixir-analyzer/issues/296

### DIFF
--- a/lib/elixir_analyzer/exercise_test/assert_call/compiler.ex
+++ b/lib/elixir_analyzer/exercise_test/assert_call/compiler.ex
@@ -80,13 +80,15 @@ defmodule ElixirAnalyzer.ExerciseTest.AssertCall.Compiler do
   """
   @spec annotate(Macro.t(), map()) :: {Macro.t(), map()}
   def annotate(node, acc) do
-    node = annotate_piped_functions(node)
+    node =
+      node
+      |> annotate_piped_functions
+      |> annotate_defdelegate
 
     acc =
       acc
       |> track_aliases(node)
       |> track_imports(node)
-      |> track_all_functions(node)
 
     cond do
       module_def?(node) -> {node, %{acc | in_module: extract_module_name(node)}}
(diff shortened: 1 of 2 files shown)


angelikatyborska:
@jiegillet I need your help here. I can't get the last two tests from this commit to pass: https://github.com/exercism/elixir-analyzer/commit/40ad1679bd4eac1634edb3fc5371e206ceedd8e6

I am very confused about how `assert_call` compiler works. Somehow it looks for the right function call twice, once in a `find` function, and then again in `indirect_call?`? Is that right? And `indirect_call?` doesn't seem to use `modules_in_scope` at all, where the alias information is stored. Why?

jiegillet:
Man, it's been a while :)

Yes, it's right, it's looking for the function call twice, but the second time seems to be very lacking. Your approach was great and should have worked.

I think we should refactor so that the first pass with `find` finds the function call leaves a tag in the metadata if it finds it, and then the second pass with `indirect_call?` can find it it's called in the right place. I will try to find the time to do that this week :)

jiegillet:
OK, I think I found a solution. It took me a while to remember how this all works.

There were two issues:
1. `track_all_functions` usually doesn't need to know about aliases, because when `function_def?` finds a match, it is added to the tree directly with a de-aliased module. However, that breaks down when the call is in a different module that is aliased.
2. `function_def?` wouldn't detect the function call in `defdelegate`, leaving it to the very limited `track_all_functions`

So I made the following changes:
1. Give `track_all_functions` the list of all aliases
2. Call `track_all_functions` in `annotate_and_find` to make sure the list of aliases is complete
3. Transform all `defdelegate` into `def` that `function_def?` can analyze

jiegillet (approved):

angelikatyborska:
Woohoo! Thank you 🙏 all the tests are passing now

Issue #350, opened 2022-11-28
HighScore feedback for module attribute is wrong


rhino232:
Essential

Use a [module attribute](https://elixir-lang.org/getting-started/module-attributes.html#as-constants) in add_player/3 and reset_score/2 as a constant for the initial score.

does not correctly recognize the solution

              

defmodule HighScore do
  @initial_score 0

jiegillet:
Hi @rhino232,
Could you copy the full solution? I believe the cause of the issue is lower than that.
Alternatively, you could share a mentoring link.
<img width="731" alt="Screen Shot 2022-11-29 at 8 51 48" src="https://user-images.githubusercontent.com/16929078/204404947-34fcd5c9-faa2-4393-9991-4717c77de6db.png">

We had a similar problem in #341, with a fix waiting to be merged in #345. Maybe it's the same problem.

rhino232:
Hi Jie

here is my mentoring link
https://exercism.org/mentoring/external_requests/eb9a48fcd2b042cf8e21463bb0065668

regards


Am Di., 29. Nov. 2022 um 00:53 Uhr schrieb Jie ***@***.***>:

jiegillet:
Thank you, I looked into it and it was a quite interesting find.
The piece that triggers the comment is 

```elixir
  def reset_score(scores, name) do
    scores
    |> Map.replace!(name, @initial_score)
  rescue
      _e in KeyError ->
        scores
        |> Map.put_new(name, @initial_score)
  end
```

I had never seen a `rescue` block` without a `try` before. But it seems to work fine, it's just rarely used syntax.
The reason why it triggered the comment is that the analyzer only expects exactly one block in a `def`, and the `rescue` block is a second one.
We should be able to fix this (it might take some time).

In the mean time, let me share with you the expected answer:
```elixir
  def reset_score(scores, name) do
    Map.put(scores, name, @initial_score)
  end
```
In Elixir, rescuing errors is used only when strictly necessary, often there are better options that don't raise errors, like in this case `Map.put`, or functions that return ok tuples `{:ok, _}` or `{:error, _}`.

rhino232:
Thanks for the feedback!

I will adapt my solution

regards
-- andi

Am Do., 1. Dez. 2022 um 14:12 Uhr schrieb Jie ***@***.***>:

angelikatyborska:
[...]

PR #366, opened 2023-02-11
Handle def+rescue in assert_call

angelikatyborska:
This PR partially deals with https://github.com/exercism/elixir-analyzer/issues/350 by handling `def` and `defp` with `rescue` blocks in `assert_call`.

It doesn't deal with the fact that we used `form do def foo()...` in many places and those won't work with `def+rescue`. However, I don't think we should change that.

Instead, I think we should add a common check to all learning exercises that don't require any rescuing that warns students that using raising and rescuing errors for flow control is an antipattern and might result in other analyzer feedback being invalid.


jiegillet (approved):
Yes, this is very nice, thank you.

And I agree with you, making a common check warning about unnecessary raising is a good idea.

PR #369, opened 2023-02-25
Add common check for not using `rescue`

angelikatyborska:
Comment PR: https://github.com/exercism/website-copy/pull/2250

This PR treats all usages of `rescue` as bad, except for the two concept exercises that teach how to use `rescue.`.

### DIFF
--- a/lib/elixir_analyzer/exercise_test/common_checks/no_rescue.ex
+++ b/lib/elixir_analyzer/exercise_test/common_checks/no_rescue.ex
@@ -1,0 +1,65 @@
+defmodule ElixirAnalyzer.ExerciseTest.CommonChecks.NoRescue do
+  @moduledoc """
+  This is an exercise analyzer extension module used for common tests looking for the usage of `rescue`
+  """
+
+  alias ElixirAnalyzer.Constants
+
+  defmacro __using__(_opts) do
+    quote do
+      feature Constants.solution_no_rescue() do
+        type :essential
+        comment Constants.solution_no_rescue()
+        find :none
+
+        form do
+          def _ignore do
+            _ignore
+          rescue
+            _ignore
+          end
+        end
+
+        form do
+          try do
+            _ignore
+          rescue
+            _ignore
+          end
+        end
+
+        form do
+          try do
+            _ignore
+          rescue
+            _ignore
+          else
+            _ignore
+          end
+        end
+
+        form do
+          try do
+            _ignore
+          rescue
+            _ignore
+          after
+            _ignore
+          end
+        end
+
+        form do
+          try do
+            _ignore
+          rescue
+            _ignore
+          else
+            _ignore
+          after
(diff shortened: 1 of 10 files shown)


lib/elixir_analyzer/exercise_test/common_checks/no_rescue.ex
```diff
@@ -0,0 +1,180 @@
...
+        form do
+          def _ignore do
+            _ignore
+          rescue
+            _ignore
+          after
+            _ignore
+          else
+            _ignore
+          end
```
jiegillet, line 121: The try form is [missing a the `catch` clause](https://hexdocs.pm/elixir/1.13.4/Kernel.SpecialForms.html#try/1):
```elixir
try do
  do_something_that_may_fail(some_arg)
rescue
  ArgumentError ->
    IO.puts("Invalid argument given")
catch
  value ->
    IO.puts("Caught #{inspect(value)}")
else
  value ->
    IO.puts("Success! The result was #{inspect(value)}")
after
  IO.puts("This is printed regardless if it failed or succeeded")
end
```

But at this point, I'm thinking it might be easier to maintain a `check_source` on `:do` blocks followed by a `:rescue` block... This seems to pass the tests:

```elixir
defmodule ElixirAnalyzer.ExerciseTest.CommonChecks.NoRescue do
  @moduledoc """ 
  This is an exercise analyzer extension module used for common tests looking for the usage of `rescue`
  """

  alias ElixirAnalyzer.Source

  defmacro __using__(_opts) do
    quote do
      check_source Constants.solution_no_rescue() do
        type :essential
        comment Constants.solution_no_rescue()

        check(%Source{code_ast: code_ast}) do
          {_, found} =
            Macro.prewalk(code_ast, false, fn node, found ->
              case node do
                {def_op, _, [_function, [{:do, _} | rest]]} when def_op in [:def, :defp] ->
                  {node, found or List.keymember?(rest, :rescue, 0)} 

                {:try, _, [[{:do, _} | rest]]} ->
                  {node, found or List.keymember?(rest, :rescue, 0)} 

                _ ->
                  {node, found}
              end 
            end)

          not found
        end 
      end 
    end 
  end 
end
```

jiegillet (commented):
Impressive check!
I would like to propose an alternative possibly easier to maintain though.

I also doubled checked that we don't use any `rescue` in the practice exercises, so it should always possible to find a solution without it (not that it's surprising).

lib/elixir_analyzer/exercise_test/common_checks/no_rescue.ex
angelikatyborska: Sure, if you're giving me a ready made solution, I'll take it 😬

jiegillet (approved):
"Do not use a 1000 lines long quotes" 🤣🤣🤣 
Love it!

PR #372, opened 2023-03-09
Fix defdelegate in high score not finding default argument

angelikatyborska:


jiegillet (approved):
Yeah, fair enough ^^

Another approach would be to do the same thing we did in `assert_call`, change the AST of the input code to change a `defdelegate` into a `def`. Definitely more work, but a more general solution. Although it would prevent us from asserting anything about a `defdelegate` block, so maybe not a good thing.

Issue #371, opened 2023-03-05
DNA Encoding should require tail call recursion


angelikatyborska:
A big chunk of highly rated community solutions to DNA Encoding doesn't even use the concept that is taught, tail call recursion. Instead they use regular recursion. See at the end of the issue for examples.

It would be great if we found a way to detect that both `encode` and `decode` functions use tail call recursion. We would need to find all functions called from either function `encode` and `decode`, and check their definitions. At least one of the called functions has a function clause whose last call is itself, it's tail-call recursion.

## Example 1 
```elixir
defmodule DNA do
  @nucleotide_encode_table %{
    ?\s => 0b0000,
    ?\A => 0b0001,
    ?\C => 0b0010,
    ?\G => 0b0100,
    ?\T => 0b1000
  }

  @nucleotide_decode_table Enum.map(@nucleotide_encode_table, fn {k, v} -> {v, k} end)
                           |> Map.new()

  def encode_nucleotide(code_point) do
    Map.fetch!(@nucleotide_encode_table, code_point)
  end

  def decode_nucleotide(encoded_code) do
    Map.fetch!(@nucleotide_decode_table, encoded_code)
  end

  def encode(dna)
  def encode([]), do: <<>>
  def encode([head | tail]) do
    <<encode_nucleotide(head)::4, encode(tail)::bitstring>>
  end

  def decode(dna)
  def decode(<<>>), do: []
  def decode(<<head::4, tail::bitstring>>) do
    [decode_nucleotide(head) | decode(tail)]
  end
end
```

## Example 2
```elixir
defmodule DNA do
  @moduledoc false

  @type nucleotide_char :: ?\s | ?A | ?C | ?G | ?T
  @type nucleotide_bitstring :: 0 | 1 | 2 | 4 | 8
  @type dna_charlist :: [nucleotide_char()]
  @type dna_bitstring :: <<_::_*4>>

  @spec encode_nucleotide(nucleotide_char()) :: nucleotide_bitstring()
  def encode_nucleotide(?\s), do: 0b0000
  def encode_nucleotide(?A), do: 0b0001
  def encode_nucleotide(?C), do: 0b0010
  def encode_nucleotide(?G), do: 0b0100
  def encode_nucleotide(?T), do: 0b1000

  @spec decode_nucleotide(nucleotide_bitstring()) :: nucleotide_char()
  def decode_nucleotide(0b0000), do: ?\s
  def decode_nucleotide(0b0001), do: ?A
[...]

PR #374, opened 2023-03-12
Require tail-call recursion in dna-encoding

angelikatyborska:
Resolves https://github.com/exercism/elixir-analyzer/issues/371

Website copy pull request: https://github.com/exercism/website-copy/pull/2255


lib/elixir_analyzer/test_suite/dna_encoding.ex
```diff
@@ -29,4 +30,113 @@ defmodule ElixirAnalyzer.TestSuite.DNAEncoding do
...
+
+      all_recursive_functions = Enum.uniq(all_recursive_functions)
+
+      non_tail_call_recursive_functions = all_recursive_functions -- tail_call_recursive_functions
+
+      if non_tail_call_recursive_functions != [] || tail_call_recursive_functions == [] do
+        false
+      else
+        true
+      end
```
angelikatyborska, line 67: @jiegillet I'm feeling a bit silly asking this since I've been working on this project for so long... but how can I return a custom `%Comment{}` struct here? I would like to pass the two function lists to the comment template.

jiegillet: Not silly at all, I also had no idea, even though I wrote `check_source` haha.

Looking at `ElixirAnalyzer.ExerciseTest.CheckSource.Compiler` it looks like the comment is created from the parameters `name`, `comment`, `type` and `suppress_if`. But looking at `ElixirAnalyzer.Comment` it seems there is one parameter missing: the map `params`, which is what you are looking for to build a custom comment.

`params` can't be added the same way since it's produced in the function, but maybe we can change the compiler to something like (untested)

```elixir
    test_description =
      Macro.escape(fn params -> %Comment{
        name: name,
        comment: comment,
        type: type,
        suppress_if: suppress_if,
        params: params
      })  

    quote do
      (fn %Source{} = source ->
         case unquote(check_function).(source) do
           false-> {:fail, unquote(test_description).(nil)}
           true-> {:pass, unquote(test_description).(nil)}
           params when is_map(params) -> {:pass, unquote(test_description).(params)}
           _ -> raise "check must be true, false or a map of parameters for the comment"
         end 
       end).(unquote(code_source))
    end 
```

angelikatyborska: Thank you, that was close to it!

lib/elixir_analyzer/exercise_test/check_source/compiler.ex
```diff
@@ -20,10 +20,21 @@ defmodule ElixirAnalyzer.ExerciseTest.CheckSource.Compiler do
...
-           {:fail, unquote(test_description)}
+         case unquote(check_function).(source) do
+           true ->
+             {:pass, unquote(test_description)}
+
+           false ->
+             {:fail, unquote(test_description)}
+
+           {true, params} when is_map(params) ->
+             {:pass, %{unquote(test_description) | params: params}}
```
angelikatyborska, line 31: I wasn't sure if this case makes sense or how to test it - if the check passes, the results don't include the comment, right? So it doesn't matter what its params are?

joaothallis (approved):

angelikatyborska:
@jiegillet could you re-review?

lib/elixir_analyzer/exercise_test/check_source/compiler.ex
jiegillet: Yeah that's true.
If we can't test it, it's basically dead code, so we can remove it until we have a use case for it (and therefore a test).

lib/elixir_analyzer/exercise_test/check_source/compiler.ex
```diff
@@ -20,10 +20,21 @@ defmodule ElixirAnalyzer.ExerciseTest.CheckSource.Compiler do
...
+
+           {false, params} when is_map(params) ->
+             {:fail, %{unquote(test_description) | params: params}}
```
jiegillet, line 34: That's way better than my silly macro suggestion :)

lib/elixir_analyzer/test_suite/dna_encoding.ex
```diff
@@ -29,4 +30,127 @@ defmodule ElixirAnalyzer.TestSuite.DNAEncoding do
...
+          [],
+          fn node, acc -> find_all_recursive_functions(node, acc) end
+        )
+
+      all_recursive_functions = Enum.uniq(all_recursive_functions)
+
+      non_tail_call_recursive_functions = all_recursive_functions -- tail_call_recursive_functions
+
+      if non_tail_call_recursive_functions != [] || tail_call_recursive_functions == [] do
+        format_function_names = fn list ->
```
jiegillet, line 64: Any reason why this function is not a normal private function?

lib/elixir_analyzer/test_suite/dna_encoding.ex
```diff
@@ -29,4 +30,127 @@ defmodule ElixirAnalyzer.TestSuite.DNAEncoding do
...
+
+              calls when is_list(calls) ->
+                calls
+
+              call ->
+                [call]
+            end
+
+          {_, any_nested_recursive_calls?} =
+            Macro.prewalk(all_calls_in_function_def, false, fn node, acc ->
```
jiegillet, line 137: Wouldn't this do the same?
```suggestion
            Macro.prewalk(opts[:do], false, fn node, acc ->
```

lib/elixir_analyzer/test_suite/dna_encoding.ex
```diff
@@ -29,4 +30,127 @@ defmodule ElixirAnalyzer.TestSuite.DNAEncoding do
...
+      else
+        true
+      end
+    end
+  end
+
+  defp find_tail_call_recursive_functions(node, acc) do
+    acc =
+      case node do
+        {op, _meta1, [{fn_name, _meta2, args}, opts]} when op in [:def, :defp] ->
```
jiegillet, line 88: This is missing the `def  fn_name(args) when is_list(args) do ...` variant, right?

test/support/analyzer_verification/check_source.ex
```diff
@@ -61,4 +61,36 @@ defmodule ElixirAnalyzer.Support.AnalyzerVerification.CheckSource do
...
+
+      if rem(banana_count, 2) == 1 do
+        {false, %{banana_count: banana_count}}
+      else
+        {true, %{banana_count: banana_count}}
+      end
+    end
+  end
+
+  check_source "should have an even number of bananas in the code" do
```
jiegillet, line 81: Is this the same as above?

lib/elixir_analyzer/test_suite/dna_encoding.ex
```diff
@@ -29,4 +30,127 @@ defmodule ElixirAnalyzer.TestSuite.DNAEncoding do
...
+          fn_arity = length(args)
+
+          last_call_in_function_def =
+            case opts[:do] do
+              {:__block, _, calls} when is_list(calls) ->
+                List.last(calls)
```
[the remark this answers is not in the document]
jiegillet: Makes sense to me :)

jiegillet (commented):
Nice!
The exercise might need a couple more tests since checking for tail recursive functions is not trivial at all...

lib/elixir_analyzer/test_suite/dna_encoding.ex
angelikatyborska: I hope not! I don't remember

[`001d0f5` (#374)](https://github.com/exercism/elixir-analyzer/pull/374/commits/001d0f50b257f8965e78495daba47207fbb3bcf3)

test/support/analyzer_verification/check_source.ex
angelikatyborska: lol, yes! I wonder what I was trying to write there

[`ed10546` (#374)](https://github.com/exercism/elixir-analyzer/pull/374/commits/ed1054606e2f7a2b98c518b37b661d36f110ef64)

lib/elixir_analyzer/exercise_test/check_source/compiler.ex
angelikatyborska: If I remove `| params: params` from this line, I get quite a lot of errors. If I remove the whole two lines, same. The only change that keeps the tests passing would be to say `params: %{}` but I don't see how that's better than keeping `params: params`, so I'm not touching this code 🤷

lib/elixir_analyzer/test_suite/dna_encoding.ex
angelikatyborska: Oops, fooled by guards again 😢 

Added support in [`17c59dd` (#374)](https://github.com/exercism/elixir-analyzer/pull/374/commits/17c59dd5e37747f7a323ccca165352bf76b8c274)

lib/elixir_analyzer/test_suite/dna_encoding.ex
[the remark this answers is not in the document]
angelikatyborska: Done: [`5ff0a1f` (#374)](https://github.com/exercism/elixir-analyzer/pull/374/commits/5ff0a1f2e07147ae5c092cf226dc69fba92b41a6)

lib/elixir_analyzer/test_suite/dna_encoding.ex
angelikatyborska: Indeed: [`67f1b64` (#374)](https://github.com/exercism/elixir-analyzer/pull/374/commits/67f1b644b6348543dc2c3d815da3218fd8b0605c)

angelikatyborska:
@jiegillet can I get a re-review?

jiegillet (approved):
Looks good to me now!
That was a super slow response from me, again I'm sorry about that :)

angelikatyborska:
OSS cannot compete with Zelda 🎮

PR #385, opened 2023-06-24
Allow `true` to be used as an `@impl` value

### LINKED ISSUE #383: Take-A-Number Deluxe analyzer claims about missing @impl GenServer
Analyzer claims about missing `@impl GenServer` but `@impl true` should be also ok!

dabaer:
Resolves #383


dabaer:
Is it worth expanding the analyzer comment surrounding the usage of `true`  as well?

angelikatyborska:
Perfect! Yes, we could update the comment too - if you want to open a PR, it's in this file: https://github.com/exercism/website-copy/blob/main/analyzer-comments/elixir/take-a-number-deluxe/annotate_impl_genserver.md

angelikatyborska (approved):

Issue #384, opened 2023-06-22
Redundant advice on the German Sysadmin exercise


fjrivasf:
The analyzer is telling me to use the ?(char) syntax, even though that's exactly what I'm already doing.

> Elixir has a [special syntax](https://elixir-lang.org/getting-started/binaries-strings-and-char-lists.html#unicode-and-code-points) to write integers that correspond to [Unicode Code Points](https://en.wikipedia.org/wiki/Code_point): you can use a ? in front of the character literal to reveal its code point.
> 
> For example,
> 
> ?a === 97
> ?\s === 32
> ?ß === 223
> ?互 === 20114
> In the context of manipulating code points, this notation is much more readable than numbers and should always be preferred.

I'm using module attributes and comprehensions to deal with the filtering part of this problem. 
My guess is that the comprehension is being expanded at compile time, and that the analyzer is seeing raw digits even though the actual source code has none.

`@alphabet for x <- ?_..?z, x != ?`, do: x
@german_chars 'äöüß'
@german_alphabet @german_chars ++ @alphabet

  def sanitize(''), do: ''
  def sanitize([h|t]) when h not in @german_alphabet, do: sanitize t
  def sanitize([h|t]) when h in @alphabet, do: [h | sanitize t]`

angelikatyborska:
Hi! You're right that the feedback in this case is wrong, but the cause is a bit different than that.

The check is done on a string with the source code, because as you noticed, after compilation you cannot tell anymore which syntax was used to create the integer value. The check is this:

```elixir
integers = ["?ß", "?ä", "?ö", "?ü", "?_", "?a", "?z"]
Enum.all?(integers, &String.contains?(code_string, &1))
```

We knew it will be buggy when we wrote it, but it's an very important check for Elixir beginners (not knowing about the `?A` syntax is one of the common mistakes I saw in charlist related exercises when I was mentoring them) so we settled on a buggy check to have something working at least in some cases.

I'm open to ideas how to improve it, I don't have any at the moment.

fjrivasf:
Thank you for the explanation! That makes a lot of sense.

After reading the motive behind it, I don't think the current approach is wrong. 
I can't think of many other ways to check for the specific chars needed in this exercise that wouldn't pass that check and also wouldn't kill the readability of the solution. Even my approach is a little bit hacky since I'm including the underscore as part of the alphabet without further explanation.

[...]

PR #389, opened 2023-07-02
Relax german-sysadmin integer literal check

angelikatyborska:
Resolves https://github.com/exercism/elixir-analyzer/issues/384

The check is relaxed - now just one usage of `?ß`, `?ä`, `?ö`, `?ü`, `?_`, `?a`, or `?z` in the solution is enough, as long as the corresponding integer literals aren't found.

A trade-off: this will produce false-positives in scenarios in which people randomly include one of those numbers in the code, including in comments, e.g. `?ß # 223`

### DIFF
--- a/mix.lock
+++ b/mix.lock
@@ -12,6 +12,6 @@
   "metrics": {:hex, :metrics, "1.0.1", "25f094dea2cda98213cecc3aeff09e940299d950904393b2a29d191c346a8486", [:rebar3], [], "hexpm", "69b09adddc4f74a40716ae54d140f93beb0fb8978d8636eaded0c31b6f099f16"},
   "mimerl": {:hex, :mimerl, "1.2.0", "67e2d3f571088d5cfd3e550c383094b47159f3eee8ffa08e64106cdf5e981be3", [:rebar3], [], "hexpm", "f278585650aa581986264638ebf698f8bb19df297f66ad91b18910dfc6e19323"},
   "parse_trans": {:hex, :parse_trans, "3.3.1", "16328ab840cc09919bd10dab29e431da3af9e9e7e7e6f0089dd5a2d2820011d8", [:rebar3], [], "hexpm", "07cd9577885f56362d414e8c4c4e6bdf10d43a8767abb92d24cbe8b24c54888b"},
-  "ssl_verify_fun": {:hex, :ssl_verify_fun, "1.1.7", "354c321cf377240c7b8716899e182ce4890c5938111a1296add3ec74cf1715df", [], [], "hexpm", "fe4c190e8f37401d30167c8c405eda19469f34577987c76dde613e838bbc67f8"},
+  "ssl_verify_fun": {:hex, :ssl_verify_fun, "1.1.7", "354c321cf377240c7b8716899e182ce4890c5938111a1296add3ec74cf1715df", [:make, :mix, :rebar3], [], "hexpm", "fe4c190e8f37401d30167c8c405eda19469f34577987c76dde613e838bbc67f8"},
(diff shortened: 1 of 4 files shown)


dabaer:
RE the false positives, you could reduce the surface area by removing comments from the checked code:

```elixir
check(%Source{code_string: code_string}) do
  question_mark_code_points = ["?ß", "?ä", "?ö", "?ü", "?_", "?a", "?z"]
  integer_literal_code_points = ["223", "228", "246", "252", "95", "97", "122"]

  code_string = code_string |> Code.string_to_quoted!() |> Macro.to_string()

  Enum.any?(question_mark_code_points, &String.contains?(code_string, &1)) &&
    not Enum.any?(integer_literal_code_points, &String.contains?(code_string, &1))
end
```

mix.lock
```diff
@@ -12,6 +12,6 @@
   "metrics": {:hex, :metrics, "1.0.1", "25f094dea2cda98213cecc3aeff09e940299d950904393b2a29d191c346a8486", [:rebar3], [], "hexpm", "69b09adddc4f74a40716ae54d140f93beb0fb8978d8636eaded0c31b6f099f16"},
   "mimerl": {:hex, :mimerl, "1.2.0", "67e2d3f571088d5cfd3e550c383094b47159f3eee8ffa08e64106cdf5e981be3", [:rebar3], [], "hexpm", "f278585650aa581986264638ebf698f8bb19df297f66ad91b18910dfc6e19323"},
   "parse_trans": {:hex, :parse_trans, "3.3.1", "16328ab840cc09919bd10dab29e431da3af9e9e7e7e6f0089dd5a2d2820011d8", [:rebar3], [], "hexpm", "07cd9577885f56362d414e8c4c4e6bdf10d43a8767abb92d24cbe8b24c54888b"},
-  "ssl_verify_fun": {:hex, :ssl_verify_fun, "1.1.7", "354c321cf377240c7b8716899e182ce4890c5938111a1296add3ec74cf1715df", [], [], "hexpm", "fe4c190e8f37401d30167c8c405eda19469f34577987c76dde613e838bbc67f8"},
+  "ssl_verify_fun": {:hex, :ssl_verify_fun, "1.1.7", "354c321cf377240c7b8716899e182ce4890c5938111a1296add3ec74cf1715df", [:make, :mix, :rebar3], [], "hexpm", "fe4c190e8f37401d30167c8c405eda19469f34577987c76dde613e838bbc67f8"},
```
dabaer, line 15: I noticed prior to the 1.15 update `[:make, :mix, :rebar3]` was in this line, but it got removed. Weird that it is now back. I can't seem to find much information on the format of the `mix.lock` file to understand this behaviour.

angelikatyborska:
@dabaer The reason we're doing this check on the string in the first place is the same reason your suggestion wouldn't work: the `?` syntax is just syntactic sugar that gets completely stripped when producing an AST:

```elixir
iex(6)> "if x == ?A, do: ?B, else: 0"
"if x == ?A, do: ?B, else: 0"

iex(7)> v |> Code.string_to_quoted!()
{:if, [line: 1],
 [{:==, [line: 1], [{:x, [line: 1], nil}, 65]}, [do: 66, else: 0]]}

iex(8)> v |> Macro.to_string()
"if(x == 65) do\n  66\nelse\n  0\nend"
```

jiegillet (approved):
I'm good with this.

Maybe one way to reduce the false positives a little bit would be to forbid the use of any of the `integer_literal_code_points` except if their question mark equivalent is also used.

```elixir
# OK
?ß -> ~c"ss"
# Not OK
223 -> ~c"ss"
# Still OK because both are used
 ?ß -> ~c"ss" # 223
```

angelikatyborska:
@jie like this [`7839ac0` (#389)](https://github.com/exercism/elixir-analyzer/pull/389/commits/7839ac0bf2dfe1d50a35e3457f9fec1a06538039) ?

jiegillet:
@angelikatyborska yes, looks good! (you called the wrong jie in the process though haha)

angelikatyborska:
Hahaha, oh no 🙈 that explains why you didn't react

Issue #368, opened 2023-02-25
Extend `CommonChecks.DebugFunctions` to look for calls to `dbg`


angelikatyborska:
- Extend the module `ElixirAnalyzer.ExerciseTest.CommonChecks.DebugFunctions` to also find calls to `dbg()`
- Add tests
- Update comment https://github.com/exercism/website-copy/blob/main/analyzer-comments/elixir/solution/debug_functions.md

elfenlaid:
Replying to https://mas.to/@angelikatyborska/110878322509896396

I'd like to give it a try, if you don't mind 🙏

angelikatyborska:
@elfenlaid Awesome! Ping me here if you have any questions.

angelikatyborska:
Done and deployed 🙌 

![Screenshot 2023-08-15 at 15 56 22](https://github.com/exercism/elixir-analyzer/assets/7852553/5f2f329c-d8c1-452c-87e1-5578a5ea7f30)

elfenlaid:
Awesome, thanks 🙇

PR #391, opened 2023-08-15
Extend common checks to also look for `dbg` calls

elfenlaid:
Part of https://github.com/exercism/elixir-analyzer/issues/368

The changes add `Kernel.dbg` to the list of checked functions

### DIFF
--- a/lib/elixir_analyzer/exercise_test/common_checks/debug_functions.ex
+++ b/lib/elixir_analyzer/exercise_test/common_checks/debug_functions.ex
@@ -12,6 +12,12 @@ defmodule ElixirAnalyzer.ExerciseTest.CommonChecks.DebugFunctions do
         comment Constants.solution_debug_functions()
         called_fn module: IO, name: :inspect
       end
+
+      assert_no_call Constants.solution_debug_functions() do
+        type :informative
+        comment Constants.solution_debug_functions()
+        called_fn module: Kernel, name: :dbg
+      end
     end
   end
 end
--- a/test/elixir_analyzer/exercise_test/common_checks_test.exs
+++ b/test/elixir_analyzer/exercise_test/common_checks_test.exs
@@ -282,13 +282,6 @@ defmodule ElixirAnalyzer.ExerciseTest.CommonChecksTest do
           def foo(name) do
             Debug.dbg(name)
           end
-        end,
-        defmodule MyModule do
-          import Kernel, only: [dbg: 1]
-
-          def foo(name) do
-            dbg(name)
-          end
         end
       ]
     end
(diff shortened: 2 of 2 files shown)


test/elixir_analyzer/exercise_test/common_checks_test.exs
```diff
@@ -266,6 +266,32 @@ defmodule ElixirAnalyzer.ExerciseTest.CommonChecksTest do
...
+        end,
+        defmodule MyModule do
+          alias Kernel, as: Debug
+
+          def foo(name) do
+            Debug.dbg(name)
+          end
+        end,
+        defmodule MyModule do
+          import Kernel, only: [dbg: 1]
```
elfenlaid, line 287: Not sure how legit this test is. There are two `dbg` versions: `Kernel.dbg/2` and `Kernel.dbg/3`. I wonder what this import does, I think it imports `Kernel.dbg/2` but I might be wrong

lib/elixir_analyzer/exercise_test/common_checks/debug_functions.ex
```diff
@@ -12,6 +12,12 @@ defmodule ElixirAnalyzer.ExerciseTest.CommonChecks.DebugFunctions do
         comment Constants.solution_debug_functions()
         called_fn module: IO, name: :inspect
       end
+
+      assert_no_call Constants.solution_debug_functions() do
+        type :informative
+        comment Constants.solution_debug_functions()
+        called_fn module: Kernel, name: :dbg
```
angelikatyborska, line 19: Perfect 👍

test/elixir_analyzer/exercise_test/common_checks_test.exs
angelikatyborska: There are three versions of `Kernel.dbg`, with 0, 1, and 2 arguments. You can see that in the docs for [`Kernel.dbg/2`](https://hexdocs.pm/elixir/Kernel.html#dbg/2) - technically the docs only list a 2 argument variant, but both arguments are defined as optional, so we know for sure 0 and 1 arity versions must exist too 🙂.

`dbg/3` is from another module, it's from `Macro`: https://hexdocs.pm/elixir/Macro.html#dbg/3 and it's not related to this ticket.

Anyway,  calling `import Kernel, only: [dbg: 1]` has a really fun side effect - it un-imports all of the other functions and macros imported from Kernel 😁 for example:

```elixir
iex()> import Kernel, only: [dbg: 1]
Kernel
iex()> 2 + 3
error: undefined function +/2 (there is no such import)
  iex:4

** (CompileError) cannot compile code (errors have been logged)
```

This error happens because the `+/2` macro is normally available because the whole `Kernel` module is usually imported by default.

What does all of this mean for this PR? We don't really need a test for importing `Kernel.dbg`. The first two test cases are enough 👍

elfenlaid: Oh my, thanks for the followup 🙇

angelikatyborska (approved):
Thank you ❤️ 🧡 💛

PR #399, opened 2023-10-12
Test fix: do not depend on map order

jiegillet:
There was an [update in OTP26](https://www.erlang.org/blog/otp-26-highlights/#changed-ordering-of-atom-keys): 
> In OTP 26, as an optimization for certain map operations, such as maps:from_list/1, maps with atom keys are now sorted in a different order. The new order is undefined and may change between different invocations of the Erlang VM.

We had some tests that relied on that order. I'm not sure I got them all since the order can vary.

The smoke test also failed on my machine because the compiler warnings were in a different order, but not on CI. I'm not sure it's related or not, but I also sorted the warnings.


angelikatyborska:
Does that mean that the warnings will be now in alphabetical order? That's a bit weird for debugging

jiegillet:
Not, it's not alphabetical order. The warnings are tuples `{filepath, line, warning}`, so they will be sorted by file, line, then warning. Seems like the best ordering possible :)

angelikatyborska:
Ah, that's perfect then 🙂

angelikatyborska (approved):

PR #400, opened 2023-10-13
Closes #398 - Fixed newsletter analysis to allow guard clause

### LINKED ISSUE #398: Newsletter in Elixir track has incorrect automated feedback
In submitting to the Elixir track my solution to the Newsletter exercise, I received the following automated feedback. I don't understand why it marks as `Essential` a code suggestion that my submitted code already has.

Screenshot 2023-10-12 at 08 12 55

octowombat:
Following [discussions](https://github.com/exercism/elixir-analyzer/issues/398) with @jiegillet and an upstream code fix.

### DIFF
--- a/lib/elixir_analyzer/test_suite/newsletter.ex
+++ b/lib/elixir_analyzer/test_suite/newsletter.ex
@@ -19,6 +19,14 @@ defmodule ElixirAnalyzer.TestSuite.Newsletter do
         end
       end
     end
+
+    form do
+      def close_log(_ignore) when _ignore do
+        _block_ends_with do
+          :ok
+        end
+      end
+    end
   end
 
   feature "use IO.puts in log_sent_email rather than IO.write as last call" do
@@ -30,6 +38,14 @@ defmodule ElixirAnalyzer.TestSuite.Newsletter do
       def log_sent_email(_ignore, _ignore) do
         _block_ends_with do
           IO.write(_ignore, _ignore)
+        end
+      end
+    end
+
+    form do
+      def log_sent_email(_ignore, _ignore) when _ignore do
+        _block_ends_with do
+          :ok
         end
       end
     end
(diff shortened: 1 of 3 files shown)


lib/elixir_analyzer/test_suite/newsletter.ex
```diff
@@ -107,6 +139,14 @@ defmodule ElixirAnalyzer.TestSuite.Newsletter do
         end
       end
     end
+
+    form do
+      def open_log(_ignore) when _ignore do
+        _block_includes do
+          File.open!(_ignore, [:write])
```
jiegillet, line 146: Let's also include all of the variations we had before (with and without `File`, with and without `!`). I know they are not all super likely, but we might as well.

test_data/two_fer/imperfect_solution/lib/two_fer.ex
```diff
@@ -1,5 +1,4 @@
 defmodule TwoFer do
-	@someUnusedModuleAttribute 1
```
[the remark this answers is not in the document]
octowombat: I did change it as it caused an unused compiler warning and my setup locally stops in that case!!

I will revert the change and retest. Thanks

lib/elixir_analyzer/test_suite/newsletter.ex
octowombat: I didn't remove any of the existing variants though.

jiegillet: I meant let's duplicate all four existing variants and add the `when _ignore`, not just for `File.open!`.

Right now if someone uses 
```elixir
        import File
        def open_log(path) when is_binary(path) do
          open!(path, [:write])
        end
```
It will trigger a comment.

octowombat: Gotcha!

test_data/two_fer/imperfect_solution/lib/two_fer.ex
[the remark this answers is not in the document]
jiegillet: Yup, the compiler warning is on purpose, so we can test if we actually detect it :)

jiegillet (approved):
That's great, thank you so much for your contribution!

octowombat:
My pleasure

Issue #378, opened 2023-03-31
DNA Encoding should only complain about using builtin iteration constructs only in function bodies


christhekeele:
On my [first iteration](https://exercism.org/tracks/elixir/exercises/dna-encoding/iterations) of this exercise, I was warned:

> The purpose of this exercise is to teach recursion. Solve it without using list comprehensions or any of the functions from the modules `Enum`, `List`, or `Stream`.

This is a fair point--I'd missed the context of the lesson and used `for` in my solution bodies.

I rewrote it to create my [second iteration](https://exercism.org/tracks/elixir/exercises/dna-encoding/iterations), and got the same warning. This is because I'm still using `for` in the module body just to define my mapping functions, which I think is valid and does not "skip" the point of the lesson.

I'd argue we should only be checking function bodies for usage of these constructs, not the entire AST of the submitted module.

angelikatyborska:
Hi!

The link you posted is not a permanent link to your solutions, but a generic link under each student can see their own solutions. So I'm writing this response without actually seeing your solution 🙂 if you think seeing it would change my mind, please copy-paste it!

Still I am very confident that you should not be using `for` anywhere in your solution. You should really be only using tail call recursion to iterate over lists and bitstrings in this exercise. If you're using `for` for anything, you either have unnecessary code in your solution, or you haven't used tail call recursion where the exercise expected you to.

The ideal solution contains three usages of tail call recursion - once in `encode/1`, and twice in `decode/1`, because `decode/1` also needs a `reverse` function at the end. This is an ideal solution:

```elixir
defmodule DNA do
  def encode_nucleotide(?\s), do: 0b0000
  def encode_nucleotide(?A), do: 0b0001
  def encode_nucleotide(?C), do: 0b0010
  def encode_nucleotide(?G), do: 0b0100
  def encode_nucleotide(?T), do: 0b1000

  def decode_nucleotide(0b0000), do: ?\s
  def decode_nucleotide(0b0001), do: ?A
  def decode_nucleotide(0b0010), do: ?C
  def decode_nucleotide(0b0100), do: ?G
  def decode_nucleotide(0b1000), do: ?T

  def encode(dna) do
    do_encode(dna, <<>>)
  end

  defp do_encode([], acc), do: acc

  defp do_encode([n | rest], acc) do
    do_encode(rest, <<acc::bitstring, encode_nucleotide(n)::4>>)
  end

  def decode(dna) do
    do_decode(dna, [])
  end

[...]

PR #409, opened 2023-11-19
Ignore module-level calls in `assert_call`

angelikatyborska:
Resolves https://github.com/exercism/elixir-analyzer/issues/378


jiegillet (approved):
Every time I stop looking at that codebase for a while and come back I get frightened about how complex it is :)

Looks good, the only change that I would suggest would be to add a test with the code that triggered the comment in the original issue.

angelikatyborska:
> Every time I stop looking at that codebase for a while and come back I get frightened about how complex it is :)

I have the same 🙈 I'm just glad we have so many unit tests. It makes modifying the "compiler" modules less scary.

Issue #407, opened 2023-11-04
`file-sniffer` analyzer should allow string representation for `type_from_binary`


toy:
Analyzer replied with «Even though there could be other valid ways to solve this exercise, for the sake of mastering bitstrings please match the first few bytes of the argument of type_from_binary/1 with the special form <<>>.» and I expected it to be fine with strings in `<<>>` instead of only digits.
```elixir
defmodule FileSniffer do
  @exe_type "application/octet-stream"
  @bmp_type "image/bmp"
  @png_type "image/png"
  @jpg_type "image/jpg"
  @gif_type "image/gif"

  def type_from_binary(<<"\dELF", _rest::binary>>), do: @exe_type
  def type_from_binary(<<"BM", _rest::binary>>), do: @bmp_type
  def type_from_binary(<<"\x89PNG\r\n\x1A\n", _rest::binary>>), do: @png_type
  def type_from_binary(<<"\xFF\xD8\xFF", _rest::binary>>), do: @jpg_type
  def type_from_binary(<<"GIF", _rest::binary>>), do: @gif_type
end
```

jiegillet:
Hi @toy, could you please update your description with the analyzer message you got and what you expected?

angelikatyborska:
@jiegillet I've been thinking about this problem. I asked myself: what is actually important in this exercise? I think the most important part is that people use `<<>>` and `::` at least once to do pattern matching on bitstrings. I think it's less important whether the bitstring chunks get created with integers `<<0x42, 0x4D>>` or with strings `<<"BM">>`. 

I looked into the current analyzer code and discovered that, after many bug fixes after people complained their specific syntax was not accepted, the analyzer is no longer effective at detecting pattern matching reliably. E.g. this solution would pass the analyzer:

```elixir
def type_from_binary(file) do
  cond do
    String.starts_with?(file, <<0x42, 0x4D>>) ->
      "image/bmp"
  
    String.starts_with?(file, <<0x89, 0x50, 0x4E, 0x47, 0x0D, 0x0A, 0x1A, 0x0A>>) ->
      "image/png"
  
    String.starts_with?(file, <<0xFF, 0xD8, 0xFF>>) ->
      "image/jpg"
  
    String.starts_with?(file, <<0x47, 0x49, 0x46>>) ->
      "image/gif"
  
    String.starts_with?(file, <<0x7F, 0x45, 0x4C, 0x46>>) ->
      "application/octet-stream"
  end
end
```

[...]

PR #412, opened 2023-11-25
Rewrite file-sniffer analysis

angelikatyborska:
Resolves https://github.com/exercism/elixir-analyzer/issues/407

Originally this PR was called:
> Rewrite file-sniffer analysis (assert_call should also find calls in function heads)

After a deep dive, I realized that `assert_call` is already capable of finding calls in function heads.

I have initially made a wrong assumption. I thought that the `Kernel` module alone is implicitly imported into each new module. That is only partially correct. The `Kernel.SpecialForms` module is also implicitly imported, with some important differences:
- You cannot modify or opt out of importing `Kernel.SpecialForms`, whereas you can modify `Kernel` imports by making them explicit and passing `:only`/`:except` options
- In the AST, usages of Kernel functions can be distinguished from local calls, but usages of `Kernel.SpecialForms` macros cannot. They look exactly like usages of local function/macros.

I was also confused which module provides which macro, I thought `<<>>` and `::` comes from `Kernel`, but it coems from `Kernel.SpecialForms`

**TL;DR**: we can use `assert_call called_fn name: :<<>>` to check for calls of special forms with no problems and no modifications necessary. It is not allowed to overwrite those special forms anyway, so you won't have a conflicting local function.

Related PR: https://github.com/exercism/website-copy/pull/2306


angelikatyborska:
@jiegillet could you review this one?

test/elixir_analyzer/test_suite/file_sniffer_test.exs
```diff
@@ -147,40 +153,90 @@ defmodule ElixirAnalyzer.ExerciseTest.FileSnifferTest do
...
+
+        def type_from_binary(<<127, 69, 76, 70, _::binary>>), do: type_from_extension("exe")
+        def type_from_binary(file), do: type_from_extension(file)
       end
     ]
   end
 
-  test_exercise_analysis "doesn't use pattern matching",
-    comments: [Constants.file_sniffer_use_pattern_matching()] do
+  test_exercise_analysis "doesn't use <<>> nor ::",
```
jiegillet, line 210: The comment should be triggered if the solution only uses one of the forms right? Should we add tests to check that?

jiegillet (approved):
Wonderful, it's nice to see that `assert_call` was already powerful enough to handle that, thank you for adding the new tests to highlight the behavior.

Sorry that I took so long to review, I still thought that `assert_call` needed to change, and I was slowly gathering motivation to deep dive back into it, but it seems I never had enough haha. I'm glad you reminded me :)

The code looks great, I just left a suggestion to add a test or two, but it's just a formality.

jiegillet:
GitHub doesn't seem to let me re-run the external tests, I can't find the button, I don't know why.

test/elixir_analyzer/test_suite/file_sniffer_test.exs
angelikatyborska: I added a test with just `<<>>` but I don't think I can add a test with just `::`. It's used for defining binary sizes, so you can't really use it without using the binary special form...

angelikatyborska:
>  I can't find the button, I don't know why.

You can only rerun the tests if the latest test run isn't too old... I think maybe 2 or 4 weeks.

jiegillet:
Ah that makes sense :)

PR #432, opened 2024-07-19
Allow any default parameter in two-fer

### LINKED ISSUE #431: `two-fer` doesn't recognize default argument unless it is `"you"`
Analyzer emits the following message for the code below.
```
Recommended
Use a default argument for this solution.
```
```
defmodule TwoFer do
  @doc """
  Two-fer or 2-fer is short for two for one. One for you and one for me.
  """
  @spec two_fer(String.t()) :: String.t()
  def two_fer(name \\ "") # This is a function head with a default argument.
  def two_fer(""), do: "One for you, one for me."
  def two_fer(name) when is_binary(name), do: "One for #{name}, one for me."
end
```

skeltoac:
Fixes #431

### DIFF
--- a/lib/elixir_analyzer/test_suite/two_fer.ex
+++ b/lib/elixir_analyzer/test_suite/two_fer.ex
@@ -45,19 +45,19 @@ defmodule ElixirAnalyzer.TestSuite.TwoFer do
 
     # function header
     form do
-      def two_fer(_ignore \\ "you")
+      def two_fer(_ignore \\ _ignore)
     end
 
     # function without a guard and with a do block
     form do
-      def two_fer(_ignore \\ "you") do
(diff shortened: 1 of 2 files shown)


jiegillet:
Hi @skeltoac, thank you for the PR.
While I believe that using `"you"` as the default parameter is the best way of solving the exercise, it's true that the comment is not accurate, so I'm OK with your proposed change.

That being said, could you:
1. Fix the failing test in`test/elixir_analyzer/test_suite/two_fer_test.exs` (basically remove the one on line 94)
2. Add a new test  `test_exercise_analysis "accepts any value as the default parameter"` (with the line 94 test, the intended solution and maybe an extra one with an atom or something)

jiegillet (approved):
Nice! Thanks a lot :)

PR #441, opened 2024-12-21
Update to Elixir 1.18

jiegillet:
Nothing much to say, no warning.


jiegillet:
@angelikatyborska I'm having trouble with this one. Thank god for smoke tests.

Basically, `bin/build.sh` which ultimately calls `mix escript.build` doesn't work as expected. It produces a 2.7K file versus 2.7M file on main. And when you run it, it gives 
```
ERROR! Failed to start Elixir.
error: {error,{elixir,{"no such file or directory","elixir.app"}}}
```
I'm guessing because Elixir is not embeded in the escript as it should. The changelog only mentions one change about escript, [this PR](https://github.com/elixir-lang/elixir/pull/13730), but it should not be relevant.

I found a way to make it work locally by doing first `mix compile`, then `ERL_LIBS=~/.asdf/installs/elixir/1.18.0-otp-27/lib:_build/dev/lib bin/elixir_analyzer slug . .`, and I'm sure we could make it work on the docker file, but the escript should work.

jiegillet:
Update, I found the culprit: `consolidate_protocols: false` is the culprit, and I think it's definitely related to that PR.
The problem is that it was there because `# Turn off protocol consolidation to avoid warning in analyzed code` so if I remove it, we get a warning in the smoke test:

```
warning: the String.Chars protocol has already been consolidated, an implementation for Clock has no effect. If you want to implement protocols after compilation or during tests, check the "Consolidation" section in the Protocol module documentation
  lib/clock.ex:51
```

angelikatyborska:
The change in behavior looks unexpected to be, so I reported it as a bug here: https://github.com/elixir-lang/elixir/issues/14096

jiegillet:
Fixed and merged in two hours. Wow. Now we wait for 1.18.1

jiegillet:
Great, seems to work fine with 1.18. I'll update the rest of the repo with the same docker image.

Issue #340, opened 2022-10-13
Allow for alternative erlang solutions in `captains-log`


aedwardg:
Given that the main focus of this exercise is to use erlang libaries, it feels strange to receive analyzer errors for the following two solutions:

`random_ship_registry/0` (prompts to use `Enum.random/1`):
```elixir
def random_ship_registry_number() do
  digits = :rand.uniform(9000) + 999

  "NCC-#{digits}"
end
```

`format_stardate/1` (prompts to use `:io_lib.format("~.1f", [stardate]) |> to_string()`):
```elixir
def format_stardate(stardate) do
  stardate
  |> :erlang.float_to_binary([{:decimals, 1}])
end
```

Would it make sense to have the analyzer allow one or both of these alternate solutions, since they are still using erlang?

jiegillet:
Hi @aedwardg, Sorry for the slow reply and thank you for opening this issue.

The exercise is teaching two concepts, randomness and using erlang libraries, this explains why we are nudging the students towards `Enum.random`. However, upon closer look, we are delivering the message

> Use `:rand.uniform` instead of `Enum.random` in the `random_stardate` function. `:rand.uniform` allows you to generate a random float directly, as opposed to `Enum.random` that only allows you to generate random integers.

which is not fully accurate, [according to the docs](https://www.erlang.org/doc/man/rand.html#uniform-1) `:rand.uniform(n)` does indeed return an integer between 1 and `n`. In the spirit of learning about Elixir `Enum.random` function, I think I would argue that using `Enum.random(0..n)` is more explicit about the range, so it is preferable. What would you think about changing the message to

> Use `:rand.uniform` instead of `Enum.random` in the `random_stardate` function. `:rand.uniform/1` allows you to generate a random integer greater or equal to 1, but `Enum.random` takes a range argument, which is more explicit and permissive.

considering the second message, I think this is a good suggestion. `:io_lib.format` is more powerful, but `:erlang.float_to_binary` is enough to satisfy all the requirements and should therefore be allowed. As a matter of fact, we should probably accept any solution that uses erlang, not just these two.

Would you like to help with implementing those changes?

aedwardg:
[...]

PR #455, opened 2025-09-16
Added erlang ast test in `captains-log`

FraSanga:
As described in issue #340, I modified the file by attempting to perform the `check` with the AST.

I am uncertain if this is correct, but the tests I added also work with the examples I created.

As soon as I get the go-ahead, I'll also make a PR on website-copy.
I hope this has been helpful. 😄


FraSanga:
It should be fixed now.
Every time Windows messes up the formatting at the end of the line.

lib/elixir_analyzer/constants.ex
```diff
@@ -76,7 +76,7 @@ defmodule ElixirAnalyzer.Constants do
     captains_log_do_not_use_enum_random: "elixir.captains-log.do_not_use_enum_random",
     captains_log_do_not_use_rand_uniform_real: "elixir.captains-log.do_not_use_rand_uniform_real",
     captains_log_use_rand_uniform: "elixir.captains-log.use_rand_uniform",
-    captains_log_use_io_lib: "elixir.captains-log.use_io_lib",
+    captains_log_use_erlang: "elixir.captains-log.use_erlang",
```
jiegillet: This value is a path to a file in [this repo](https://github.com/exercism/website-copy) that contains the text of the analyzer comment, so you need to create a new file over there for the analyzer to work. Could you do the appropriate change over there, link to the PR in this PR's description and request a review from me?

lib/elixir_analyzer/test_suite/captains_log.ex
```diff
@@ -43,10 +44,30 @@ defmodule ElixirAnalyzer.TestSuite.CaptainsLog do
...
+      {_, erlang?} =
+        Macro.prewalk(code_ast, false, fn node, acc ->
+          case node do
+            # usage :io_lib.format/2
+            {{:., _, [:io_lib, :format]}, _, _} ->
+              {node, true}
+
+            # usage :erlang.function_name/arity
+            # matches any function call from :erlang module
+            {{:., _, [:erlang, _]}, _, _} ->
```
jiegillet: Could you add at least one test in `test/elixir_analyzer/test_suite/captains_log_test.exs` that use an `:erlang` function?

jiegillet (commented):
Sorry about the wait. I think I was waiting for you to address my second point, and then lost track of the PR altogether.

> 2. The tests in `test_data` are used as smoke tests for the CI, we don't typically add new ones in there unless we have a good reason. The tests you added in `test/elixir_analyzer/test_suite/captains_log_test.exs` should be enough.

This still stands, could you revert the changes to `test_data`?

Other than that, I've looked at your changes and left some comments.

lib/elixir_analyzer/constants.ex
FraSanga: https://github.com/exercism/website-copy/pull/2395

I cannot ask anyone for a review on the repo.

FraSanga:
I should have sorted everything out. My only concern with the first approach is that if a user makes a mistake, both the comments from :io_lib and :erlang will be displayed, I suppose.

jiegillet (commented):
Thanks for the quick follow up, I didn't take 5 months to reply this time :)

We are on the right track, but your concern is warranted, so let's address it.

jiegillet (approved):
Great job, thank you very much for your contribution.

I jumped in and added a quick fix, because I merged the sister PR a little too early and I didn't want to leave a broken state for too long, sorry about that.

PR #466, opened 2026-06-12
Update to Elixir 1.20

jiegillet:
Before merging, https://github.com/exercism/elixir/pull/1635 needs to be merged and pulled here

We get a bunch of compiler warnings because of all the macro shenanigans that this repo is the master of. I tried fixing them, but I gave up pretty quickly. I suspect that these warnings will be fixable once type annotations are a thing though, not all hope is lost.


LuuOW (commented):
Technical audit: code patterns and functional logic verified for system integrity.

angelikatyborska:
Interesting. I can see a lot of warnings now, but all of them stem from this code, right?

https://github.com/exercism/elixir-analyzer/blob/c9c1ed5684cc4947bfbaafdff652562b6397a9af/lib/elixir_analyzer/exercise_test/check_source/compiler.ex#L23-L38

As long as the tests pass, I'm fine with ignoring the warnings for now. We can address them when we actually need to work on the logic of the project again, or wait for improvements in the type checker.

angelikatyborska (approved):

jiegillet:
That’s right, it’s all from that piece of code. From what I can see, the logic is sound, it’s just that some modules provably can’t use all branches, but all branches are used by some module so we can’t remove any. 
I’m sure we could find a workaround but I think that in principle type annotations should be able to resolve these kinds of issues, and that seems like the best approach so I’m willing to wait it out. 

Thanks for reviewing and merging all of these :)

